PR #18, opened 2024-05-24
Update to the main branch

timfelle:
Big updates to the workflows, which should now be useful in general.

Submodules have been removed, we now handle external dependencies in the setup script.

Some general updates to source files and examples.


timfelle:
This PR would not follow the correct setup for issuing a new release.

PR #30, opened 2024-08-19
Mesh modularity

timfelle:
Updates to the meshing script, we now support the following file formats:

- `.jou`, journal files from Cubit
- `.geo`, gmesh geometry files
- `.exo`, Exodus files, not yet tested with multiblock files.
- `.re2`, just since we are often doing this as part of the others.

### DIFF
--- a/scripts/meshing.sh
+++ b/scripts/meshing.sh
@@ -1,0 +1,159 @@
+# ============================================================================ #
+# Functions for meshing using Cubit, Exodus, and Nek5000
+# ===========================================================================
+#
+# This script contains functions to create and convert meshes using Cubit,
+# Exodus, and Nek5000. The functions are used in the `mesh.sh` script to
+# automate the meshing process.
+
+# ============================================================================ #
+# Cubit Journal to neko binary mesh
+
+function jou2nbin() {
+    set -e
+
+    find_cubit
+    find_exo2nek
+    find_rea2nbin
+
+    input_file=$1
+    input_name=$(basename ${1%.*})
+
+    # Construct the mesh using Cubit
+    $cubit -nojournal -nographics -batch -noecho $1
+
+    if [[ $(find ./ -name "*.exo" | wc -l) -lt 1 ]]; then
+        printf >&2 "\n\e[4mError:\e[0m\n"
+        printf >&2 "  %-10s %-67s\n" "Cubit:" "Exodus not created: $input_file"
+        exit 1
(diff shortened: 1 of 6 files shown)


scripts/meshing.sh
```diff
@@ -0,0 +1,159 @@
...
+
+    # Convert the mesh to Neko mesh format
+    $rea2nbin ${mesh_file%.*}.re2 $input_name.nmsh
+
+    if [[ $(find ./ -name "*.nmsh" | wc -l) -lt 1 ]]; then
+        printf >&2 "\n\e[4mError:\e[0m\n"
+        printf >&2 "  %-10s %-67s\n" "rea2nbin:" "re2 not created: $input_file"
+        exit 1
+    fi
+}
```
gloopydoop, line 159: yo, it's going to take me a while to go through this whole PR, so sorry if you don't get any feedback in the upcoming days.
I've just scrolled to the gmsh section because it's the one I'm the most familiar with. 

There's one problem I foresee here is: periodic boundary conditions. 
`Nek5000` and of course now `neko` treat periodic boundary conditions as internal conditions and then topologically link them to the other side.
Because of this, if you want to prescribe periodic boundaries, you will have to specify the boundaries in `gmsh2nek` and assign a normal vector connecting them.

I'm not exactly sure how we can handle this in a user friendly "one-script-fits-all" like you're trying to achieve here. I suppose we could either:
- allow additional user scripts accompanying the gmsh file that indicate the surfaces/normal vector
- or maybe for this PR specifically it's ok to write in the documentation somewhere that periodic boundaries are not currently supported.

Again, I'm a bit busy today so it will take me a while to get around to this PR, but if you could provide me with some meshes in these formats (if they're light I guess you could include them in `neko-top`, otherwise just send them to me directly) I can test out these scripts on my end and check that things seem reasonable.

timfelle: for now I think the solution would be to document that we don't support the periodic boundaries. In regards to sending meshes. You have `.geo` files you can use, the `.jou` goes from `.jou` to `.exo` to `.re2` to `.nmsh` so there you have all the others. If you want to test them sepperately, you can use the `-k` flag for `mesh.sh` and then use the intermediaries as inputs.

timfelle: @gloopydoop  checkout the added documentation

gloopydoop (approved):
good stuff!

PR #43, opened 2024-10-08
Fix job submission and improve logging

timfelle:
This pull request includes several commits that fix issues related to job submission and improve logging and messages to the user. The changes ensure that the job submission works correctly on different clusters, such as MareNostrum 5 and DTU. The submission process has been updated to use batch scripts associated with the desired cluster, and the output of the job is stored in the results folder after completion. Additionally, the logging and messages to the user have been improved for better clarity and understanding.


documentation/examples.md
```diff
@@ -55,11 +78,18 @@ are required:
 
 ## Case files and meshes.
 
-When running an example a link is made to the local data folder if it exists.
+When running an example a link is made to the folders `data` and `data_local`
+both sitting at the root of Neko-TOP. The `data` folder contain some official
+meshes used in the predefined examples. The `data_local` allow the user to save
+meshes to a folder accessible to all examples to reduce need for redundancy.
```
gloopydoop, line 84: This is actually my first time reading this documentation.
I think it would more intuitive to have either:
- an empty `data_local` sitting in `neko-top` by default and gitignore everything inside it.
- Have none of the examples refer to `data_local` in the case files

gloopydoop (approved):
Examples run fine and documentation looks nice!

documentation/examples.md
timfelle: Well the goal is to have the meshing tool save the preprocessed meshes into the `data_local` folder, which is why the examples refer to it. We want to supply the smallest possible files to reproduce our examples, and sometimes that is a gmsh or jou file without the binary mesh.

I dont think git support commiting an empty folder, it would make no sense as it is for controlling files.

PR #48, opened 2024-10-16
masks

gloopydoop:
Here I've added masks based on `point_zones` for both the design indicator and source terms.
- volume constraint is now based on the optimization domain
- we have some basic masking functionality in `mask_ops` but I intend to extend this in the future for more complex functionality (and GPUs)
- this is a big ***warning*** for Abbas (currently implementing the `optimizer_t`) I've just masked the sensitivity. This is not the correct way of handling the interface with MMA, because the $N$ is wrong. You should use the masks on initialization of MMA such that only those within the mask are flagged as design variables.

### DIFF
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -105,9 +105,11 @@ program usrneko
      ! Abbas, don't just mask the sensitivity like I'm doing here, make sure
      ! the only design variables entering MMA are those within the mask.
      ! This way you get the correct N etc.
-     call mask_exterior_const(&
-     problem%volume_constraint%sensitivity_to_coefficient, &
-     design%optimization_domain, 0.0_rp)
+     if (design%if_mask) then
+        call mask_exterior_const(&
+        problem%volume_constraint%sensitivity_to_coefficient, &
+        design%optimization_domain, 0.0_rp)
+     end if
 
      ! now we have the optimizer act on the design field.
 
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -105,6 +105,10 @@ program usrneko
      ! Abbas, don't just mask the sensitivity like I'm doing here, make sure
      ! the only design variables entering MMA are those within the mask.
      ! This way you get the correct N etc.
+
+     ! Look into the `masked_red_copy` function that Martin implemented.
+     ! That function will copy from one array to another, but the target 
+     ! only have the size of the mask, not the full size.
      if (design%if_mask) then
         call mask_exterior_const(&
         problem%volume_constraint%sensitivity_to_coefficient, &
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -21,6 +21,7 @@ program usrneko
   use mma, only: mma_t
   use neko_ext, only: reset
   use math, only: copy, cmult
+  use mask_ops, only: mask_exterior_const
 
 
   !> a problem type
@@ -100,6 +101,14 @@ program usrneko
      ! in this case it's MMA so we need gradient information
      call problem%compute_sensitivity()
 
+     ! TODO
+     ! Abbas, don't just mask the sensitivity like I'm doing here, make sure
+     ! the only design variables entering MMA are those within the mask.
+     ! This way you get the correct N etc.
+     call mask_exterior_const(&
+     problem%volume_constraint%sensitivity_to_coefficient, &
+     design%optimization_domain, 0.0_rp)
+
      ! now we have the optimizer act on the design field.
 
      fval(1) = problem%volume_constraint%objective_function_value
@@ -112,6 +121,7 @@ program usrneko
        ! TODO
        ! I'm CERTAIN the mass matrix comes in here, but I need to sit down
        ! with a pen and paper
+
 
        ! TODO
        ! for heuristic reasons, it's important to rescale the dfdx etc so
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -404,7 +404,7 @@ contains
     ! test something in the passive scalar.
     if (this%if_mask) then
        call mask_exterior_const(this%sensitivity, this%optimization_domain, &
-       0.0_rp)
+            0.0_rp)
     end if
 
     call neko_scratch_registry%relinquish_field(temp_indices)
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -280,7 +280,7 @@ contains
     ! Initialize the mask
     if (this%if_mask) then
        this%optimization_domain => &
-       neko_point_zone_registry%get_point_zone(optimization_domain_zone_name)
+            neko_point_zone_registry%get_point_zone(optimization_domain_zone_name)
     end if
 
 
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -393,12 +393,9 @@ contains
     this%filtered_design)
 
     ! TODO
-    ! oh oh... I don't know what to do here...
-    ! I'm just following the vibe of "everything in the adjoint happens in
-    ! reverse". So I'm masking at the end.
-    ! This could be wrong, talk to the big dogs.
+    ! DELETE THIS LATER
     !
-    ! Also, depending on how Abbas writes the interface for the optimization
+    ! When Abbas writes the interface for the optimization
     ! module this may be a moot point, because we would only really collect
     ! the sensitivity of the design variables inside the mask.
     !
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -197,10 +197,6 @@ module topopt_design
      type(field_t) :: filtered_design
 
 
-     !
-     ! TODO
-     ! you also had lots of masks etc that was a nice idea,
-     ! but we'll cross that bridge later
      !> A mask indicating the optimization domain
      class(point_zone_t), pointer :: optimization_domain
      !> A logical if we're restricting the optimization domain
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -316,7 +316,7 @@ contains
     ! if we mask first then filter, at least all the boundaries will be smooth.
     if (this%if_mask) then
        call mask_exterior_const(this%design_indicator, &
-       this%optimization_domain, 0.0_rp)
+            this%optimization_domain, 0.0_rp)
     end if
 
     ! TODO
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -41,6 +41,9 @@ module topopt_design
   use coefs, only: coef_t
   use scratch_registry, only: neko_scratch_registry
   use fld_file_output, only: fld_file_output_t
+  use point_zone_registry, only: neko_point_zone_registry
+  use point_zone, only: point_zone_t
+  use mask_ops, only: mask_exterior_const
 
   implicit none
   private
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -390,7 +390,7 @@ contains
     this%filtered_design)
 
     call this%filter%apply_backward(this%sensitivity, dF_dfiltered_design, &
-    this%filtered_design)
+         this%filtered_design)
 
     ! TODO
     ! DELETE THIS LATER
--- a/sources/designs/topopt_design.f90
+++ b/sources/designs/topopt_design.f90
@@ -102,7 +102,9 @@ module topopt_design
   use linear_mapping, only: linear_mapping_t
   use RAMP_mapping, only: RAMP_mapping_t
   use PDE_filter, only: PDE_filter_t
+  use mask_ops, only: mask_exterior_const
   !use design_variable, only: design_variable_t
+  use point_zone, only: point_zone_t
   implicit none
   private
 
@@ -199,6 +201,10 @@ module topopt_design
      ! TODO
      ! you also had lots of masks etc that was a nice idea,
      ! but we'll cross that bridge later
+     !> A mask indicating the optimization domain
+     class(point_zone_t), pointer :: optimization_domain
+     !> A logical if we're restricting the optimization domain
+     logical :: if_mask
 
      ! TODO
      ! you also had logicals for convergence etc,
@@ -244,6 +250,7 @@ contains
     class(topopt_design_t), target, intent(inout) :: this
     type(json_file), intent(inout) :: json
     type(coef_t), intent(inout) :: coef
+    character(len=:), allocatable :: optimization_domain_zone_name
     integer :: n, i
     ! init the fields
     call this%design_indicator%init(coef%dof, "design_indicator")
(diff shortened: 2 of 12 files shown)


sources/neko_ext/mask_ops.f90
```diff
@@ -0,0 +1,136 @@
...
+          work%x(mask%mask(i), 1, 1, 1) = fld%x(mask%mask(i), 1, 1, 1)
+       end do
+    end if
+
+    ! copy over
+    call field_copy(fld, work)
+
+    call neko_scratch_registry%relinquish_field(temp_indices)
+
+  end subroutine mask_exterior_const
```
timfelle: Did you look at math?
https://github.com/ExtremeFLOW/neko/blob/5c4e1d1cd02b499ee77f59ac88805ce18fd760e1/src/math/math.f90#L285-L296

sources/neko_ext/mask_ops.f90
```diff
@@ -0,0 +1,136 @@
...
+!
+!> Some common Masking operations we may need
+!! TODO
+!! I know I got outvoted on defining a mask as a field of booleans...
+!! You were right, `point_zones` are convenient, but I don't like that the
+!! `invert` is defined on initialization (maybe I'm misunderstanding)
+!! So we can chose either the interior or the exterior at the beginning.
+!!
+!! There's probably a smarter way of doing this, but if you want to zero the
+!! interior vs the exterior, one will involve some awkward memory shifting.
```
timfelle: If you want to do things in two different masks you should create two masks.
So create one for the interior if you need the interior, and one for the
exterior if you need that.

sources/neko_ext/mask_ops.f90
```diff
@@ -0,0 +1,136 @@
...
+    ! copy the fld in the masked region
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call neko_error('GPU not supported for masks yet')
+    else
+       do i = 1, mask%size
+          fld%x(mask%mask(i), 1, 1, 1) = const 
+       end do
+    end if
+
+  end subroutine mask_interior_const
```
timfelle: or even `sources/neko_ext/math_ext.f90`

sources/objectives/minimum_dissipation_objective_function.f90
```diff
@@ -226,6 +233,10 @@ subroutine minimum_dissipation_objective_function_compute(this, design, fluid)
     call field_addcol3(objective_field, wo2, wo2)
     call field_addcol3(objective_field, wo3, wo3)
 
+    if (this%if_mask) then
+       call mask_exterior_const(objective_field, this%mask, 0.0_rp)
+    end if
```
timfelle: We should probably not spend the time computing all those gradients above if you
just 0 it anyway.

sources/neko_ext/mask_ops.f90
gloopydoop: Would we now not be in the position that we're 4x heavier than a field of booleans... since combined we're effectively storing the same number of GLL pts, but now with integers?

sources/neko_ext/mask_ops.f90
gloopydoop: Oh these are nice!

sources/neko_ext/mask_ops.f90
timfelle: Not quite, since many compilers will store a logical in a 4 byte variable on the ram.
But also, yes memory wise that is a thing, but what about compute. How often do you need to run through these masks? It is much faster to run through 10% of our domain than adding a branchpoint at every single iteration in our loop.

timfelle: What is it you want to do to everything outside the mask?

gloopydoop: well the majority of mask applications end up being used to force some GLL pts to zero. So the applications in the loops could be
`fld(i) = real(mask(i),kind=rp)*fld(i)`

In terms of how often we use mask, they're used in filtering and mapping, which is a few times per optimization iteration.

In time time dependent optimization problems we generally have objectives of the form:
$\int_{time} \int_{space} f(x)$
Where the time integral is done by accumulating at each timestep.
So either you: 
- take a field, time integrate the whole field, then at the end you mask and integrate in space.
- at each time step, mask, integrate in space, then the time integral becomes an accumulation of that scalar.

In the later case, that's an application every timestep, but also a `glsc2` every timestep.
In the former, it's one at the end, and one `glsc2` at the end, but you require an extra field for the accumulation.

gloopydoop: > What is it you want to do to everything outside the mask?

Take for instance, an optimization domain defined on $\Omega_{O} \subset \Omega$.
Then the design variables live on the interior, so calculating volumes etc require the inside.
But to "mask" the design requires forcing everything that's outside of $\Omega_{O}$ to 0 or 1, depending on if we want everything outside of $\Omega_{O}$ to be solid or fluid

timfelle: Oh, you intend to use these "masks" to define the design itself and not just the feasible domain of our design?

timfelle: > well the majority of mask applications end up being used to force some GLL pts to zero. So the applications in the loops could be `fld(i) = real(mask(i),kind=rp)*fld(i)`

Why would you update GLL points in the first place that you don't want to update. In any case, you are probably holding your information in a temporary array, so you can either set everything outside to zero, or just only copy everything inside to another variable.

gloopydoop: > Oh, you intend to use these "masks" to define the design itself and not just the feasible domain of our design?

No I mean feasible domain of our design. I'm saying if you take a boolean subtraction between $\Omega$ and $\Omega_{O}$ So everything in the computational domain that is not in the optimization domain. That still has to be either solid or fluid. Which is why you need to outside.

timfelle: Lets discuss it tomorrow.

From what I see we don't need to explicitly do anything on just the exterior.

gloopydoop: Sure!
My morning to totally filled with meetings and seminar, so in the arvo?

Maybe a final thing to be thinking about, in terms of explicitly doing things on the interior vs the exterior, is where we define these variables. For instance, I agree that $\rho(\mathbf{x})$ should be defined with $\mathbf{x} \in \Omega_O$, but when we start mapping etc, something like $\chi (\mathbf{x})$ I'm saying $\mathbf{x} \in \Omega$.

It gets worse when we think about the PDE filters. $\tilde{\rho}(\mathbf{x})$, do we assert this is applicable for $\mathbf{x} \in \Omega_O$ or $\mathbf{x} \in \Omega$. We could interpreter this different way  (I drew a little picture to explain)

In circumstances such as these, "masking" a field isn't applying an operation only to the GLL pts inside the mask. It's applying an operation to the entire field, and only retaining what's inside the mask.

See you tomorrow!

PXL_20241016_144904151

gloopydoop: > > well the majority of mask applications end up being used to force some GLL pts to zero. So the applications in the loops could be `fld(i) = real(mask(i),kind=rp)*fld(i)`
> 
> Why would you update GLL points in the first place that you don't want to update. In any case, you are probably holding your information in a temporary array, so you can either set everything outside to zero, or just only copy everything inside to another variable.

well yeh, I mean that's exactly what `mask_exterior_const` is doing.

sources/objectives/minimum_dissipation_objective_function.f90
gloopydoop: yeh that's a fair point, but the alternative is writing a `masked_grad`. 
If we wanted to make this super extensible we would write a whole `masked_operators.f90`, I suppose in a similar spirit to what you were doing with `math_ext`.
Perhaps it's best to stick this this on the todo list and save it for another PR.

sources/neko_ext/mask_ops.f90
gloopydoop: Bit of a shame on that front. The mask format is different to those used in `point_zone`, the standard is:
`mask Mask array of length m+1, where mask(0) =m`

gloopydoop:
- Moved glsc2_mask to math_ext 
- used glsc2_mask in minimum dissipation objective

sources/neko_ext/mask_ops.f90
```diff
@@ -0,0 +1,116 @@
...
+    ! fill background fld
+    call field_copy(work, background)
+
+    ! copy the fld in the masked region
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call neko_error('GPU not supported for masks yet') 
+    else
+       do i = 1, mask%size
+          work%x(mask%mask(i), 1, 1, 1) = fld%x(mask%mask(i), 1, 1, 1)
+       end do
```
timfelle, line 107:
```suggestion
       call device_masked_copy(work%x_d, fld%x_d, mask%mask, &
            fld%size(), mask%size())
    else
       call masked_copy(work%x, fld%x, mask%mask, fld%size(), mask%size())
```

sources/objectives/minimum_dissipation_objective_function.f90
```diff
@@ -228,7 +236,12 @@ subroutine minimum_dissipation_objective_function_compute(this, design, fluid)
 
     ! integrate the field
     n = wo1%size()
-    this%dissipation = glsc2(objective_field%x, fluid%C_Xh%b, n)
+    if (this%if_mask) then
+       this%dissipation = glsc2_mask(objective_field%x, fluid%C_Xh%b, &
+       n, this%mask%mask, this%mask%size)
```
timfelle:
```suggestion
            n, this%mask%mask, this%mask%size)
```

sources/objectives/minimum_dissipation_objective_function.f90
```diff
@@ -239,7 +252,12 @@ subroutine minimum_dissipation_objective_function_compute(this, design, fluid)
        call field_col3(objective_field, fluid%u, design%brinkman_amplitude)
        call field_addcol3(objective_field, fluid%v, design%brinkman_amplitude)
        call field_addcol3(objective_field, fluid%w, design%brinkman_amplitude)
-       this%lube_value = glsc2(objective_field%x, fluid%C_Xh%b, n)
+       if (this%if_mask) then
+          this%lube_value = glsc2_mask(objective_field%x, fluid%C_Xh%b, &
+          n, this%mask%mask, this%mask%size)
```
timfelle:
```suggestion
               n, this%mask%mask, this%mask%size)
```

sources/objectives/volume_constraint.f90
```diff
@@ -65,6 +65,8 @@ module volume_constraint
   use math, only : glsc2
   use field_math, only: field_rone, field_cmult
   use topopt_design, only: topopt_design_t
+  use mask_ops, only:  mask_exterior_const
```
timfelle:
```suggestion
  use mask_ops, only: mask_exterior_const
```

sources/objectives/volume_constraint.f90
```diff
@@ -149,27 +179,28 @@ subroutine volume_constraint_compute(this, design, fluid)
...
     ! TODO
     ! in the future we should be using the mapped design varaible
     !corresponding to this constraint!!!
-    this%volume = glsc2(design%design_indicator%x, fluid%c_xh%B, n)
+    if (design%if_mask) then
+       if (neko_bcknd_device .eq. 1) then
+          call neko_error('GPU not supported volume constraint')
+       else
+          this%volume = glsc2_mask(design%design_indicator%x, fluid%c_xh%B, &
+          n, this%mask%mask, this%mask%size)
```
timfelle:
```suggestion
               n, this%mask%mask, this%mask%size)
```

sources/designs/topopt_design.f90
```diff
@@ -267,6 +270,55 @@ subroutine topopt_design_init(this, json, coef)
...
+    ! the initial design field we'll be reading things properly from the JSON.
+    ! call json_get(json, 'name', optimization_domain_zone_name)
+    ! Right now, I'm hardcoding the name of the point zone.
+    this%if_mask = .true.
+    optimization_domain_zone_name = "optimization_domain"
+
+    ! Initialize the mask
+    if (this%if_mask) then
+       this%optimization_domain => &
+       neko_point_zone_registry%get_point_zone(optimization_domain_zone_name)
```
timfelle:
```suggestion
            neko_point_zone_registry%get_point_zone(optimization_domain_zone_name)
```

sources/designs/topopt_design.f90
```diff
@@ -267,6 +270,55 @@ subroutine topopt_design_init(this, json, coef)
...
+    !
+    ! We can always change this decision later, but I'm going with (1)
+    ! mask first, then filter.
+    ! The reason being, one of the purposes of the filtering is to avoid sharp
+    ! interfaces, if we filter first then mask there's a chance we have a sharp
+    ! interface on the boundary of the optimization domain.
+    ! if we mask first then filter, at least all the boundaries will be smooth.
+    if (this%if_mask) then
+       call mask_exterior_const(this%design_indicator, &
+       this%optimization_domain, 0.0_rp)
```
timfelle:
```suggestion
            this%optimization_domain, 0.0_rp)
```

sources/designs/topopt_design.f90
```diff
@@ -334,6 +392,21 @@ subroutine topopt_design_map_backward(this, df_dchi)
     call this%filter%apply_backward(this%sensitivity, dF_dfiltered_design, &
     this%filtered_design)
```
timfelle:
```suggestion
         this%filtered_design)
```

sources/designs/topopt_design.f90
```diff
@@ -334,6 +392,21 @@ subroutine topopt_design_map_backward(this, df_dchi)
...
+    ! When Abbas writes the interface for the optimization
+    ! module this may be a moot point, because we would only really collect
+    ! the sensitivity of the design variables inside the mask.
+    !
+    ! Note for Abbas,
+    ! I'm NOT doing this because I'm too lazy and I just need masks so I can
+    ! test something in the passive scalar.
+    if (this%if_mask) then
+       call mask_exterior_const(this%sensitivity, this%optimization_domain, &
+       0.0_rp)
```
timfelle:
```suggestion
            0.0_rp)
```

sources/objectives/volume_constraint.f90
```diff
@@ -129,7 +136,30 @@ subroutine volume_constraint_init(this, design, fluid, adjoint)
...
+       design%optimization_domain%name)
+
+       ! calculate the volume of the optimization domain
+       call neko_scratch_registry%request_field(work , temp_indices(1))
+       call field_rone(work)
+       if (neko_bcknd_device .eq. 1) then
+          call neko_error('GPU not supported volume constraint')
+       else
+          this%volume_domain = glsc2_mask(work%x, fluid%c_xh%B, &
+          n, this%mask%mask, this%mask%size)
```
timfelle:
```suggestion
               n, this%mask%mask, this%mask%size)
```

timfelle (changes requested):

sources/neko_ext/mask_ops.f90
gloopydoop: I tried this at first, the indexing Martin was using for `masked_copy` is different to the indexing used in `point_zone`.

`@param mask Mask array of length m+1, where mask(0) =m`

you end up with segfaults :(

timfelle: Did you provide `m+1` as a size input or did you provide `m` as input?

timfelle: I have opened an issue about it on Neko. It is definitely a mistake to have two ways to define a mask in the same codebase.

examples/rugby_ball/driver.f90
```diff
@@ -100,6 +101,16 @@ program usrneko
      ! in this case it's MMA so we need gradient information
      call problem%compute_sensitivity()
 
+     ! TODO
+     ! Abbas, don't just mask the sensitivity like I'm doing here, make sure
+     ! the only design variables entering MMA are those within the mask.
+     ! This way you get the correct N etc.
```
timfelle, line 107:
```suggestion
     ! TODO
     ! Abbas, don't just mask the sensitivity like I'm doing here, make sure
     ! the only design variables entering MMA are those within the mask.
     ! This way you get the correct N etc.

     ! Look into the `masked_red_copy` function that Martin implemented.
     ! That function will copy from one array to another, but the target 
     ! only have the size of the mask, not the full size.
```

sources/neko_ext/mask_ops.f90
timfelle: Lets wait with this for now.

timfelle (dismissed):

timfelle (approved):

PR #52, opened 2024-10-21
Handling of the problem type

timfelle:
This PR exposes some of the work we are doing in improving the problem type.
The hope is to make it relatively generic and minimize which thing need to be specified by a derived type.

This PR will primarily touch:

- How an optimization problem is defined (problem.f90, and derived types)
- How the problem interact with a "design" or solution. (design.f90, and detived types)
- Representation and handling of constraints and objectives.


sources/optimizer/mma_optimizer.f90
```diff
@@ -28,177 +32,184 @@ module mma_optimizer
...
-      logical :: auto_scale
-  contains
-      ! Override the deferred methods
-      procedure :: init => mma_optimizer_init
-      procedure :: run => mma_optimizer_run
-      procedure :: free => mma_optimizer_free
-
-      procedure, pass(this) :: run_ss => mma_optimizer_run_steady_state_prob
+     type(mma_t) :: mma
+     type(topopt_design_t), pointer :: design
```
AbbasMoosaV, line 36: Why you added a pointer to design in the mma_optimizer_t?
We discussed that at some point and we agreed on providing the design as an input to the mma_optimizer_t and remove the pointer. right?

sources/optimizer/mma_optimizer.f90
```diff
@@ -28,177 +32,184 @@ module mma_optimizer
...
 
   !> Initialize the MMA optimizer, associate it with a specific problem
-  subroutine mma_optimizer_init(this, prob)
+  subroutine mma_optimizer_init(this, problem, design)
     class(mma_optimizer_t), intent(inout) :: this
-    class(problem_t), intent(inout) :: prob
+    class(problem_t), intent(inout) :: problem
+    type(topopt_design_t), target, intent(in) :: design
+
+    this%design => design
```
AbbasMoosaV, line 63: I think we don't need a pointer to the design.

sources/optimizer/mma.f90
```diff
@@ -144,6 +144,7 @@ subroutine mma_init_json(this, x, n, json, scale, auto_scale)
     ! ----------------------------------------------------- !
     class(mma_t), intent(inout) :: this
     integer, intent(in) :: n
+    integer, intent(in) :: m
     real(kind=rp), intent(in), dimension(n) :: x
```
AbbasMoosaV, line 148: I suggest to define the design variable x as a vector_t.
It is easier to handle that for device backend later on.

sources/optimizer/mma_optimizer.f90
timfelle: For compatibility.
The `mma_optimizer_run_steady_state_prob` was accessing the
design from the problem. So in order to allow the code to function with minimal
changes outside the problem type itself I added the pointer here.

This PR should not modify the optimizer more than simple things to keep it
compatible with the problem type. Any redesign of the optimizer will come in a
separate PR.

sources/optimizer/mma.f90
timfelle: Sure, but that does not belong in this PR.

AbbasMoosaV (approved):

PR #53, opened 2024-10-21
Enforce LF line endings and clean up code

timfelle:
This pull request enforces the use of LF line endings in the repository and cleans up the code by removing unused variables and turning off a warning that should not be an error. These changes improve code consistency and maintainability.


CMakeLists.txt
gloopydoop, line 1: Ok so if I understand this correctly, we have two ways to compile, `Debug` and `Release`, and `Debug` is super pedantic and `Release` is chill. But on `Debug`, things won't compile if we have things like unused variables etc.

So after this PR, when github does the compilation check it does it on `Debug` so everything needs to be up to the pedantic code style before it can be pulled in.

In principle I don't HATE this, but it is even more extreme than the linter stuff.
The reason I don't hate it is because at the beginning of this project, one of the things I wanted to learn was better coding style, so at the end of the day this PR will be moving in that direction.

Maybe before I accept the PR, what's the intended workflow here?
So you do all your hacky development locally, and it will still compile.
And then what would you write as the flag after `./setup` to compile with debug to clean things up?

timfelle: Yes you can do whatever you want on your own machine and on the branch as you push it, this will just remind you of all the temporary stuff you added for the development once you submit the PR. This way we keep the `develop` branch on github as clean as possible.

Well the linter is purely aestetic, this is actually more functional. The checks we do here ensure we don't use variables before they are initialized for example. It is worth noting that GNU Fortran them selves mention that these checks are no guarantee for complying with the Fortran standard and having good code, but it is a step in the right direction.

A good deal of the checks here can lead to undefined behavior, segfaults and other hard to debug errors. For example there were multiple places where a variable was used before it had a value. Luckily gfortran often defaults to 0, but there is no guarantee and we could have very unpredictable outcomes, sometimes finishing with weird results, sometimes doing the right thing, and sometimes not completing at all.

You can keep track of this on your own machine by applying the cmake variable to the setup script
```bash
CMAKE_VARIABLES="-DCMAKE_BUILD_TYPE=Debug" ./setup.sh
```

timfelle: And btw, no Debug is not pedantic and Release is not chill. Debug will spent the time to ensure things are in order and tell you if you made a mistake. Release will crunch ahead, expecting your code to be perfect and let you deal with whatever comes out at the end, segfault, correct results or whatever in between.

gloopydoop: > Yes you can do whatever you want on your own machine and on the branch as you push it, this will just remind you of all the temporary stuff you added for the development once you submit the PR. This way we keep the `develop` branch on github as clean as possible.
> 
> Well the linter is purely aestetic, this is actually more functional. The checks we do here ensure we don't use variables before they are initialized for example. It is worth noting that GNU Fortran them selves mention that these checks are no guarantee for complying with the Fortran standard and having good code, but it is a step in the right direction.
> 
> A good deal of the checks here can lead to undefined behavior, segfaults and other hard to debug errors. For example there were multiple places where a variable was used before it had a value. Luckily gfortran often defaults to 0, but there is no guarantee and we could have very unpredictable outcomes, sometimes finishing with weird results, sometimes doing the right thing, and sometimes not completing at all.
> 
> You can keep track of this on your own machine by applying the cmake variable to the setup script
> 
> ```shell
> CMAKE_VARIABLES="-DCMAKE_BUILD_TYPE=Debug" ./setup.sh
> ```

Yeh this all makes sense, and I think it's a good idea! 
As I said, always happy to learn better coding practice, and I have introduce countless bugs in the past for things like uninitialized variables :sweat_smile:.
So I am very in favour of this PR. 

Maybe one last thing, because I guess Abbas (or any other people who may end up contributing) may not be reading this chat (and for me who has a horrible memory and will end up forgetting :sweat_smile:) if you can add something to readme about this:
```shell
CMAKE_VARIABLES="-DCMAKE_BUILD_TYPE=Debug" ./setup.sh
```

that would be great

timfelle: I have added a `CONTRIBUTING.md` file. It should contain what we need. hopefully it will also be supported by github, so that a link to it will be shown when one wish to open a pr

gloopydoop (approved):

CMakeLists.txt
timfelle: @gloopydoop If you are happy with the responses of the conversation please "Resolve Conversation". Otherwise github does not know we are done and ready to complete the PR.

PR #56, opened 2024-10-25
Passive scalar

gloopydoop:
don't review this yet I'm checking the linting!!


sources/objectives/minimum_dissipation_objective_function.f90
```diff
@@ -255,22 +262,25 @@ end subroutine minimum_dissipation_objective_function_compute
 
   !> compute the sensitivity of the objective function with respect to $\chi$
```
timfelle:
```suggestion
  !> compute the sensitivity of the objective function with respect to \f$\chi\f$
```
Doxygen does not directly understand `$` to mean a function. You have to tell it by using `\f$` instead.

PR #57, opened 2024-10-29
Update interface to new Neko interface

timfelle:
This pull request updates the interface to the new Neko interface.


gloopydoop (approved):
Sure, this seems reasonable for compilation/execution.

But while we have this PR up I just want to add one comment: 
Really, we should have this whole thing commented out, or throw an error saying it's not currently functional. I've actually had a number of discussions with people back in the `Nek5000` days about `vol_flow` in the adjoint solve. I'm of the opinion that the adjoint `vol_flow` should enforce a bulk velocity of 0. I'm not sure if that was specific to the case we were working on, and the adjoint derivation we came up with, or if this should be the case under all circumstances.

But for now I guess we should tread carefully using `vol_flow` in the adjoint.

PR #67, opened 2024-12-09
Fix interfaces to Neko

timfelle:


gloopydoop (commented):
These changes seem reasonable, but maybe while this PR is up it's a good time to be discussing the `init_from_json` in the first place with respect to these adjoint source terms.

So obviously a lot of the structure was copied for the normal source terms, but their use always comes by appending them with the source term handler.

So at no point do we actually use the `init_from_json`.
So perhaps now is a good time to clean that out entirely and somehow indicate this?

timfelle (commented):
are these adjoint source terms not an adjoint counterpart to a non-adjoint source term. In that case the adjoint source term must be able to be initialized with the same interface as the forward counterpart to make sure they are correctly mirroring eachother.

gloopydoop (approved):

PR #71, opened 2025-01-08
Feature/optimizer

AbbasMoosaV:
- An optimizer type is added to improve the structure of the topology optimization solver.
- MMA.f90 is updated to read the required parameters from the json case file.
- Two different ways of scaling the objective and its derivatives is considered.
- The rugby_ball example is updated to rugby_ball_optimizerobj to be use the optimizer type.
- The old rugby_ball example is no longer supported due to changes in MMA.f90.
to do list:
- The ...%free() functions needs to be checked and updated.


sources/mma/mma_optimizer.f90
```diff
@@ -0,0 +1,236 @@
...
+  use csv_file, only: csv_file_t
+  use matrix, only: matrix_t
+
+  implicit none
+  private
+  public :: mma_optimizer_t
+
+  ! Concrete type for MMA optimizer
+  type, extends(optimizer_t) :: mma_optimizer_t
+
```
timfelle: This should not contain a `mma_t` it should be the `mma_t`.

sources/mma/mma_optimizer.f90
```diff
@@ -0,0 +1,236 @@
...
+      type(mma_t) :: mma
+
+      !Pointer to the specific problem_type(steady_state_problem_t in this case)
+      type(steady_state_problem_t), pointer :: steady_state_prob => null()
```
timfelle: please avoid pointing to the problem itself. The problem should be an input when needed.

sources/problems/steady_state_problem.f90
```diff
@@ -285,11 +285,11 @@ subroutine steady_state_problem_compute(this)
...
-    print *, 'OBJECTIVE FUNCTION', &
-         this%objective_function%objective_function_value
-    print *, 'VOLUME CONSTRAINT', &
-         this%volume_constraint%objective_function_value, &
-         this%volume_constraint%volume
+    !print *, 'OBJECTIVE FUNCTION', &
+    !     this%objective_function%objective_function_value
+    !print *, 'VOLUME CONSTRAINT', &
+    !     this%volume_constraint%objective_function_value, &
+    !     this%volume_constraint%volume
```
timfelle: Avoid non-changes in other files to keep the pr simple
```suggestion
    print *, 'OBJECTIVE FUNCTION', &
         this%objective_function%objective_function_value
    print *, 'VOLUME CONSTRAINT', &
         this%volume_constraint%objective_function_value, &
         this%volume_constraint%volume
```

sources/optimizer.f90
timfelle, line 1: Please place this in a `optimizer` folder and place `mma` with it.

sources/CMakeLists.txt
```diff
@@ -17,7 +17,8 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/neko_ext)
 target_sources(neko-top
     PUBLIC
         ${CMAKE_CURRENT_SOURCE_DIR}/neko_top.f90
-
+	${CMAKE_CURRENT_SOURCE_DIR}/optimizer.f90
+
```
timfelle: Please do not mix formatting of files. The file use 4 spaces for indentation
levels. This list of files are indented 2 levels.
You added a new line with tabulated indentation and a single level, which does
not match the existing code.
```suggestion
        ${CMAKE_CURRENT_SOURCE_DIR}/optimizer.f90

```

sources/mma/mma.f90
```diff
@@ -129,7 +129,7 @@ end subroutine mma_KKT_cpu
 
 contains
 
-  subroutine mma_init_json(this, x, n, m, a0, a, c, d, xmin, xmax, json)
+  subroutine mma_init_json(this, x, n, json, scale, auto_scale)!, m, a0, a, c, d, xmin, xmax, json)
```
timfelle: Please remove the old inputs if they are not needed.

timfelle (changes requested):
If the old rugby ball is not supported then replace it with the new one instead of adding a new example and leaving a broken example.

sources/mma/mma_optimizer.f90
AbbasMoosaV: I don't quite understand that!
So here we extend the optimizer_t to mma_optimizer_t which contains mma_t.
But if I understand this comment correctly, you want to remove the mma_optimizer_t and modify the mma.f90 such that

`type, public, extends(optimizer_t) :: mma_t`

The issue is that mma_t does not need a pointer to problem. 
So I have added mma_optimizer_t to be the middle guy containing both mma_t and a pointer to the problem (or having the problem as an **input** if we don't want to have a **pointer** to the problem).
Moreover, the mma_t does not have an optimization loop (run() function). Therefore,  I have added the run() function to the mma_optimizer_t which calls the mma_t update() functiton in an  optimization loop.

I think this way, the structure would be cleaner compare to removing the mma_optimizer_t and do all these stuff in the mma_t directly as an extended optimizer_t.
What do you think?

sources/optimizer/optimizer.f90
```diff
@@ -0,0 +1,61 @@
...
+
+        !> Scaling fval and dfdx.
+        !! Note that the values are not updated but they are scaled when passed 
+        !! to the optimizer.
+        !! (if auto_scale then fval=scale else fval=scale*fval)
+        !! When auto_scale is true, we use an adaptable scale for
+        !! fval and dfdx in every iteration (variable scale factors)
+        
+        real(kind=rp) :: scale
+        logical :: auto_scale
```
timfelle: This should not be here.
If MMA needs some scaling, then it should be implemented in MMA

sources/optimizer/mma.f90
```diff
@@ -429,7 +455,7 @@ subroutine mma_free(this)
   end subroutine mma_free
 
   ! ========================================================================== !
-  ! Getters and setters
+  !>Getters and setters
```
timfelle: This is not a doxygen documentation for a function or an object. It is a regular comment.
```suggestion
  ! Getters and setters
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+    type(matrix_t) :: opt_data
+
+
+    max_iter = this%mma%get_max_iter()
+    ! call MPI_Comm_rank(neko_comm, rank, ierr)
+    call MPI_Allreduce(this%mma%get_n(), nglobal, 1, &
+        MPI_INTEGER, mpi_sum, neko_comm, ierr)
+
+    !>initializing the scaling factor
+    scalingfactor = 1_rp
```
```suggestion
    scalingfactor = 1.0_rp
```

sources/mma/mma_optimizer.f90
timfelle: Well there is a point in making `mma_t` as close to the original method as possible such that it can be reused elsewhere, without having to untangle it from the interfaces we add here.

sources/optimizer/mma_optimizer.f90
timfelle, line 1: Why are you saving the optimization data twice?
Both to the optimization_data.txt and optimization_data.csv
Please pick one and dump the other

On top of this you are also printing the same values to the standard output, leading to 3 copies of the information.

sources/designs/topopt_design.f90
```diff
@@ -205,7 +205,7 @@ subroutine topopt_design_init(this, json, coef)
     do i = 1, n
        if (sqrt((this%design_indicator%dof%x(i,1,1,1) - 0.5_rp)**2 + &
             (this%design_indicator%dof%y(i,1,1,1) &
-            - 0.5_rp)**2) .lt. 0.25_rp) then
+            - 0.5_rp)**2) .lt. 0.16_rp) then
```
timfelle: Why is this part of an MMA pull request? What does this have to do with MMA?

sources/problems/steady_state_problem.f90
```diff
@@ -286,10 +286,10 @@ subroutine steady_state_problem_compute(this)
     call this%objective_function%compute(this%design, this%C%fluid)
     call this%volume_constraint%compute(this%design, this%C%fluid)
     print *, 'OBJECTIVE FUNCTION', &
-         this%objective_function%objective_function_value
+        this%objective_function%objective_function_value
     print *, 'VOLUME CONSTRAINT', &
-         this%volume_constraint%objective_function_value, &
-         this%volume_constraint%volume
+        this%volume_constraint%objective_function_value, &
+        this%volume_constraint%volume
```
timfelle: There is no need for this to be in the PR, especially since it moves from
following our code style, to not doing so.
```suggestion
         this%objective_function%objective_function_value
    print *, 'VOLUME CONSTRAINT', &
         this%volume_constraint%objective_function_value, &
         this%volume_constraint%volume
```

sources/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/neko_ext)
 target_sources(neko-top
     PUBLIC
         ${CMAKE_CURRENT_SOURCE_DIR}/neko_top.f90
-
+
```
timfelle: Extra whitespace, please don't add that.
```suggestion

```

examples/rugby_ball/driver.f90
```diff
@@ -1,46 +1,20 @@
...
-  use scratch_registry, only : neko_scratch_registry
   use num_types, only : rp
-  use field_math, only: field_rzero, field_rone, field_cmult
-  use volume_constraint, only: volume_constraint_t
-  use fld_file_output, only : fld_file_output_t
-  use steady_state_problem, only : steady_state_problem_t
-  use mma, only: mma_t
-  use neko_ext, only: reset
-  use math, only: copy, cmult
+
```
timfelle:
```suggestion

```

examples/rugby_ball/driver.f90
```diff
@@ -1,46 +1,20 @@
...
-  use neko, only: neko_init, neko_solve, neko_finalize
-  use case, only: case_t
-  use adjoint_case, only: adjoint_case_t, adjoint_init, adjoint_free
-  use simulation_adjoint, only: solve_adjoint
-
-
-  ! here are all the extra things i'll add to "problem"
-  use source_term, only: source_term_t
-  use source_term_handler, only: source_term_handler_t
+  use mma_optimizer, only : mma_optimizer_t
```
timfelle:
```suggestion
  use mma_optimizer, only : mma_optimizer_t
```

examples/rugby_ball/driver.f90
```diff
@@ -1,46 +1,20 @@
...
-  use case, only: case_t
-  use adjoint_case, only: adjoint_case_t, adjoint_init, adjoint_free
-  use simulation_adjoint, only: solve_adjoint
-
-
-  ! here are all the extra things i'll add to "problem"
-  use source_term, only: source_term_t
-  use source_term_handler, only: source_term_handler_t
+  use mma_optimizer, only : mma_optimizer_t 
+  use steady_state_problem, only : steady_state_problem_t
```
timfelle:
```suggestion
  use steady_state_problem, only: steady_state_problem_t
```

sources/optimizer/optimizer.f90
```diff
@@ -0,0 +1,61 @@
...
+    use problem, only: problem_t
+    use num_types, only : rp
+    implicit none
+    private
+
+    !> Abstract optimizer class.
+    type, abstract, public :: optimizer_t
+
+        !> Scaling fval and dfdx.
+        !! Note that the values are not updated but they are scaled when passed
```
timfelle:
```suggestion
        !! Note that the values are not updated but they are scaled when passed
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+    print *, "max_iter for the optimization loop= ", max_iter
```
```suggestion
    print *, "max_iter for the optimization loop = ", max_iter
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+    call opt_data%init(max_iter+1,6)
+
+    call prob%compute()
+    print *, "initial objective function value=" , &
```
```suggestion
    print *, "initial objective function value = " , &
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+      prob%volume_constraint%objective_function_value
+    print *, "size(prob%design%design_indicator%x)=", &
```
```suggestion
    print *, "size(prob%design%design_indicator%x) = ", &
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+      size(prob%design%design_indicator%x)
+    print *, "size(&
+      &prob%volume_constraint%sensitivity_to_coefficient%x)=",&
```
```suggestion
      &prob%volume_constraint%sensitivity_to_coefficient%x) = ",&
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -0,0 +1,239 @@
...
+      size(&
+     prob%volume_constraint%sensitivity_to_coefficient%x)
+          
+
+    !Writing the optimization data in a separate file
+    open(1368, file="optimization_data.txt", status="replace")
```
```suggestion
    open(1368, file = "optimization_data.txt", status = "replace")
```

sources/optimizer/mma.f90
```diff
@@ -176,6 +181,27 @@ subroutine mma_init_json(this, x, n, m, a0, a, c, d, xmin, xmax, json)
...
+
+
+    allocate(a(m))
+    allocate(c(m))
+    allocate(d(m))
+    a= a_const
+    c= c_const
+    d= d_const
+    xmin = xmin_const
+    xmax = xmax_const
```
timfelle:
```suggestion
    a = a_const
    c = c_const
    d = d_const
    xmin = xmin_const
    xmax = xmax_const
```

timfelle (approved):

sources/optimizer/mma_optimizer.f90
AbbasMoosaV: Well the reason for optimization_data.txt is to print the optimization info in run time. This can be helpful to check the optimization problem in run time.
At the end, we write the optimization_data.csv to save the optimization problem info cleanly in a separate file.
I wanted to have a sort of a log file for the optimization problem (.txt file in the current version) and then at the end of the simulation, properly save the optimization date in a csv file.
Which one you suggest to keep?

timfelle: Keep the run time one and remove prints. Then you have your run time information and it is saved to a file.

PR #72, opened 2025-01-23
Enhance JSON handling and driver structure improvements

timfelle:
Introduce json based initialization for all main types.

Additionally I have split the simulation out of the problem in order to simplify dependencies.

Depends on #76

### DIFF
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -48,7 +48,7 @@ program usrneko
 
   call simulation%init(parameters)
   call design%init(parameters, simulation)
-  call problem%init(parameters, simulation, design)
+  call problem%init(parameters, design, simulation)
   call optimizer%init(parameters, problem, design, simulation)
 
   ! -------------------------------------------------------------------------- !
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -1,26 +1,26 @@
 program usrneko
-  use mma_optimizer, only : mma_optimizer_t
-  use steady_state_problem, only: steady_state_problem_t
+  use simulation, only: simulation_t
   use topopt_design, only: topopt_design_t
+  use steady_state_problem, only: steady_state_problem_t
+  use mma_optimizer, only : mma_optimizer_t
 
-  use num_types, only : rp
   use json_module, only: json_file
   use utils, only: neko_error
   use json_utils_ext, only: json_read_file
 
-  use mpi_f08, only: MPI_Init, MPI_Finalize, MPI_Comm_rank, MPI_COMM_WORLD
-
+  use mpi_f08, only: MPI_Init
   implicit none
 
   ! JSON related arguments
   integer :: argc
   type(json_file) :: parameters
   character(len=256) :: parameter_file
-  real(kind=rp) :: tolerance
 
   ! MPI parameters
-  integer :: mpi_rank, ierr, max_iter
+  integer :: ierr
 
+  !> The simulation we are working with
+  type(simulation_t) :: simulation
   !> The problem type
   type(steady_state_problem_t) :: problem
   !> The design type
@@ -32,7 +32,6 @@ program usrneko
   ! Initialize the MPI environment
 
   call MPI_Init(ierr)
-  call MPI_Comm_rank(MPI_COMM_WORLD, mpi_rank, ierr)
 
   ! -------------------------------------------------------------------------- !
   ! Read the parameters file as the first terminal argument
@@ -47,27 +46,22 @@ program usrneko
   ! -------------------------------------------------------------------------- !
   ! Initialization of the components
 
-  ! init the problem (base)
-  call problem%init()
-
-  ! init the problem, with the design
-  call problem%init_design(design)
-
-  ! init the optimizer
-  call optimizer%init(problem, design)
+  call simulation%init(parameters)
+  call design%init(parameters, simulation)
+  call problem%init(parameters, simulation, design)
+  call optimizer%init(parameters, simulation, problem, design)
 
   ! -------------------------------------------------------------------------- !
   ! Execute the optimization
 
-  tolerance = 1.0e-3_rp
-  max_iter = 100
-  call optimizer%run(problem, tolerance, max_iter)
+  call optimizer%run(simulation, problem, design)
 
   ! -------------------------------------------------------------------------- !
   ! Clean up the components
 
+  call optimizer%free()
   call problem%free()
   call design%free()
-  call optimizer%free()
+  call simulation%free()
 
 end program usrneko
--- a/examples/rugby_ball/driver.f90
+++ b/examples/rugby_ball/driver.f90
@@ -2,19 +2,24 @@ program usrneko
   use mma_optimizer, only : mma_optimizer_t
   use steady_state_problem, only: steady_state_problem_t
   use topopt_design, only: topopt_design_t
+
   use num_types, only : rp
-  use mask_ops, only: mask_exterior_const
   use json_module, only: json_file
-  use mpi_f08, only: MPI_COMM_WORLD, MPI_Init, MPI_Comm_rank
   use utils, only: neko_error
   use json_utils_ext, only: json_read_file
 
+  use mpi_f08, only: MPI_Init, MPI_Finalize, MPI_Comm_rank, MPI_COMM_WORLD
+
+  implicit none
+
+  ! JSON related arguments
   integer :: argc
   type(json_file) :: parameters
   character(len=256) :: parameter_file
   real(kind=rp) :: tolerance
 
-  integer :: mpi_rank, ierr
+  ! MPI parameters
+  integer :: mpi_rank, ierr, max_iter
 
   !> The problem type
   type(steady_state_problem_t) :: problem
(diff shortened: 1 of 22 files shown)


sources/optimizer/mma_optimizer.f90
```diff
@@ -121,8 +147,11 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance)
     scalingfactor = 1.0_rp
     print *, "max_iter for the optimization loop = ", max_iter
 
-    call problem%compute(this%design)
-    call problem%compute_sensitivity(this%design)
+    call simulation%run_forward()
+    call problem%compute(design)
+
+    call simulation%run_backward()
+    call problem%compute_sensitivity(design)
```
gloopydoop, line 169: I guess I've missed a few PRs :sweat_smile:

Am I understanding this correctly:
simulation and problem have now been separated.
calling `problem%compute` is I guess just evaluating the objective and constraints?
calling `problem%compute_sensitivity` is the computing the sensitivity.

The only worry I have with this is when we move to unsteady problems, generally your objectives would be time integrals and so will your sensitivities. Meaning they would need to be accumulated  during the run forward and the run backward.

What's the plan for this?

PS. I like this big split between steady state and other stuff. I guess it gives us a lot of flexibility when we eventually define our unsteady problems etc, so I'm sure there is a solution there.

sources/problems/problem.f90
```diff
@@ -312,16 +270,15 @@ subroutine problem_read_objectives(this, json, design)
...
     do i = 1, n_objectives
-       call json_extract_item(json, path, i, objective_json)
+       call json_extract_item(parameters, path, i, objective_json)
        call json_get(objective_json, "type", type)
        call neko_log%message(type)
 
-       call objective_factory(objective, objective_json, design, &
-            this%simulation)
+       call objective_factory(objective, objective_json, design, simulation)
        call this%add_objective(objective)
```
gloopydoop, line 283: This is really cool. Maybe it was from an old PR that I've missed. But this is great

examples/rugby_ball/driver.f90
```diff
@@ -1,36 +1,69 @@
...
+  ! Read the parameters file as the first terminal argument
 
-  ! init the optimizer
-  call optimizer%init(problem, design)
+  argc = command_argument_count()
+  if (argc .lt. 1) call neko_error('Missing parameter file')
+  call get_command_argument(1, parameter_file)
 
+  ! Read the parameters file
+  parameters = json_read_file(trim(parameter_file))
```
gloopydoop, line 45: ok I guess I'm a few PR's behind, but this look great!
But I'm a little confused, so this `parameters` is which section of the JSON?

sources/problems/problem.f90
```diff
@@ -312,16 +270,15 @@ subroutine problem_read_objectives(this, json, design)
 
     ! Get the number of objectives.
     path = "optimization.objectives"
```
gloopydoop, line 272: oh I guess this answers my previous question. Ok so our parameters in the JSON exist outside the `"case"` section? correct?
so we have access to `"optimization"` and `"mma"`?
Maybe this should be part of Abbas's PR but it would be nice if `"mma"` was replaced by `"optimization algorithm"` or something to that effect, and then we select `"mma"` out of the choices of algorithms. But I guess that's for a different PR.

Do we plan to have another section in the JSON for `"design"`. Somewhere that we can pick a topology optimization design, and potentially initial material distributions etc? Sorry I feel like I'm a few PR's behind so maybe this has already been resolved, but I remember in the first version of this I hard coded the initial rugby ball shape to be a circle, but the intention for the future was to define a point zone or something to be our initial material distribution.

gloopydoop (commented):
Maybe I wasn't the best person to review this PR because I seem to be a few behind. But overall it looks really good!
Maybe in tomorrow's meeting it's better if you guys bring me up to speed with all the changes and what our plans are for the future.

But overall I think all these changes are great!

examples/rugby_ball/driver.f90
timfelle: It is the json. The entire thing

sources/optimizer/mma_optimizer.f90
timfelle: It is up to each constraint and objective to maintain themselves. Meaning, if a constraint need to accumulate information during a run, then it must initialize and append a simulation component to the simulation. That is why the objective factory function have a simulation as an input with the `inout` intent.

sources/problems/problem.f90
timfelle: It was yes

sources/problems/problem.f90
timfelle: I think the initial condition for the optimization problem should probably be part of the optimization section of the json. Additionally the MMA section, Abbas added is a temporary and will probably be part of a `solver` object since `mma` is not special.

timfelle:
I have paused this until we get the design and problem setup fixed.

examples/rugby_ball/driver.f90
```diff
@@ -1,35 +1,67 @@
...
-  ! init the problem, with the design
-  call problem%init_design(design)
+  ! -------------------------------------------------------------------------- !
+  ! Read the parameters file as the first terminal argument
 
-  ! init the optimizer
-  call optimizer%init(problem, design)
+  argc = command_argument_count()
+  if (argc .lt. 1) call neko_error('Missing parameter file')
+  call get_command_argument(1, parameter_file)
```
gloopydoop, line 41: aaaahhh this makes sense. So when we do something like `./neko rugby_ball.case` that first argument is what's being read, so the case file. ok makes sense

examples/rugby_ball/driver.f90
```diff
@@ -1,35 +1,67 @@
...
+  call optimizer_factory(optimizer, parameters, problem, design, simulation)
+
+  ! -------------------------------------------------------------------------- !
+  ! Execute the optimization
+
+  call optimizer%run(problem, design, simulation)
+
+  ! -------------------------------------------------------------------------- !
+  ! Clean up the components
+
```
gloopydoop, line 61: ok so I guess you alluded to this in our meeting on Tuesday. So prior to this we had `optimizer%init(problem, design)` 

Now we have the driver initializing everything. I suppose this begs the question, in the future are we going to introduce some larger over encompassing module containing all of this? such that the driver is simply:
```
super_module%init
super_module%compute
super_module%free
```

examples/rugby_verification/rugby_ball.case
```diff
@@ -86,6 +86,26 @@
...
+                "scale": 1.0,
+                "auto_scale": false,
+                "xmin": 0.0,
+                "xmax": 1.0,
+                "a0": 1.0,
+                "a": 0.0,
+                "c": 100.0,
+                "d": 0.0
+            }
+        },
```
gloopydoop, line 108: this is very nice

examples/rugby_verification/rugby_ball.case
```diff
@@ -86,6 +86,26 @@
         ]
     },
     "optimization": {
+        "solver": {
+            "type": "mma",
+            "max_iterations": 5,
+            "tolerance": 1.0e-3,
+            "mma": {
+                "max_iter": 100,
```
gloopydoop, line 94: I guess this was Abbas, so not so related to this PR, but we should revise some of these naming conventions

examples/rugby_verification/rugby_ball.case
```diff
@@ -86,6 +86,26 @@
...
+                "asyincr": 1.2,
+                "asydecr": 0.7,
+                "m": 1,
+                "scale": 1.0,
+                "auto_scale": false,
+                "xmin": 0.0,
+                "xmax": 1.0,
+                "a0": 1.0,
+                "a": 0.0,
+                "c": 100.0,
```
gloopydoop, line 105: again, not really this PR, but I was under the impression that `c` in mma is vector no? One for each constraint? Maybe I'll shift this comment to PR#85 @AbbasMoosaV

sources/optimizer/mma_optimizer.f90
```diff
@@ -43,83 +45,96 @@ module mma_optimizer
...
-  ! Define the optimization loop for MMA
-  subroutine mma_optimizer_run(this, problem, tolerance, max_iter)
+  !> Initialize the MMA optimizer from JSON file
+  subroutine mma_optimizer_init_from_components(this, problem, design, &
+       simulation, max_iterations, tolerance)
     class(mma_optimizer_t), intent(inout) :: this
-    class(problem_t), intent(inout) :: problem
+    class(problem_t), intent(in) :: problem
+    class(design_t), intent(in) :: design
+    type(simulation_t), intent(in) :: simulation
```
gloopydoop, line 119: At this point can we get away with an `, optional` here?

sources/problems/objectives/minimum_dissipation_objective.f90
```diff
@@ -151,10 +151,7 @@ subroutine minimum_dissipation_init_attributes(this, design, simulation, &
 
     type(adjoint_minimum_dissipation_source_term_t) :: adjoint_forcing
 
-    call this%init_base(design%size(), weight, mask_name)
-
-    ! Assign the name to objective
-    this%name = name
+    call this%init_base(name, design%size(), weight, mask_name)
```
gloopydoop, line 154: :+1:

gloopydoop (commented):
Looks good!

sources/optimizer/mma_optimizer.f90
timfelle: No, not yet. but it is getting close

examples/rugby_verification/rugby_ball.case
timfelle: It is a vector, but the value is just assigned to all values as a constant.

examples/rugby_verification/rugby_ball.case
timfelle: Possibly, but out of scope for this PR.

examples/rugby_ball/driver.f90
timfelle: Exactly. That is also what Neko does. but we are slowly diverging from their driver and making our own.

examples/rugby_ball/driver.f90
timfelle: No, the whole point is to make this more of a library than a single executable which can do everything.

examples/rugby_verification/rugby_ball.case
gloopydoop: we can change that in a later PR

gloopydoop (approved):
:+1: 
looks good, we can change the `c` comment in a different PR

PR #76, opened 2025-01-29
Re introduction of the design class

timfelle:
The bugs have now ben fixed and this is a reintroduction of the updated design types

Depends on #77


sources/optimizer/bcknd/cpu/mma_cpu.f90
```diff
@@ -243,8 +243,8 @@ subroutine mma_subsolve_dpip_cpu(this, designx)
     xsi(:) = max(1.0_rp, 1.0_rp/(x(:) - this%alpha%x(:)))
     eta(:) = max(1.0_rp, 1.0_rp/(this%beta%x(:) - x(:)))
     mu(:) = max(1.0_rp, 0.5_rp*this%c%x(:))
-
-    do while (epsi .gt. 0.9*this%epsimin)
+    iter = 0
+    do while (epsi .gt. 0.9*this%epsimin .and. iter .lt. this%max_iter)
```
AbbasMoosaV: Well `this%max_iter` is the maximum number of iterations for the main optimization loop.
I understand that it would be better to set a maximum number of iteration for this inner loop in MMA as well.
But since at the end of each iteration here we have `epsi = 0.1_rp*epsi` then we can add a limit on the number of iterations by the value of epsi.
My suggestion here is to use the following:

`do while (epsi .gt. 0.9*this%epsimin .and. epsi .gt. 1.0_rpE-12)`

Note that we usually start with epsi = 1, so even if this%epsimin=0 for some reason (which it should not be) but the additional conditions limits the number of iterations to 12.

AbbasMoosaV (approved):

PR #80, opened 2025-01-30
Test working on a merge queue

timfelle:
A few updates to the CI. Some of this is to support the merge queues by github.
The merge queue ensure multiple PR's being accepted at roughly the same time does not break the repository. Each accepted Pr is added to a queue and merged in order. This should help streamline our development, while every merge is tested before the next is added.

- [x] Develop should properly support the merge queues.
- [x] The formatting action is now functional.
- [x] Compilation check should fail on warnings, such as un-initialized variables.

### DIFF
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -43,11 +43,10 @@ jobs:
         run: |
           git fetch --unshallow origin develop
           changes=($(git diff --name-only --diff-filter=d origin/develop))
-
           echo "changed-files=${changes[@]}" >> $GITHUB_OUTPUT
 
       # ---------------------------------------------------------------------- #
-      # Format the source files and check the formatting
+      # Format the changed source files and check for changes
       # ---------------------------------------------------------------------- #
 
       - name: Format sources
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -44,14 +44,6 @@ jobs:
           git fetch --unshallow origin develop
           changes=($(git diff --name-only --diff-filter=d origin/develop))
 
-          printf "Detected fortran changes:\n"
-          for file in ${changes[@]}; do
-            if [[ ${file: -4} != ".f90" && ${file: -4} != ".F90" ]]; then
-              continue
-            fi
-            printf "\t- $file\n"
-          done
-
           echo "changed-files=${changes[@]}" >> $GITHUB_OUTPUT
 
       # ---------------------------------------------------------------------- #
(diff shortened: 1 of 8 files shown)


.github/workflows/check_formatting.yml
```diff
@@ -35,15 +35,32 @@ jobs:
...
+
+      # ---------------------------------------------------------------------- #
+      # Format the changed source files and check for changes
       # ---------------------------------------------------------------------- #
 
       - name: Format sources
         env:
-          FINDENT_FLAGS: ${{ vars.FINDENT_FLAGS }}
+          FINDENT_FLAGS: "-Rr -i2 -d3 -f3 -s3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0"
+
```
gloopydoop, line 55: I guess this is the crux yeh?

I have to say, at first I was really against this strict formatting, but I've changed my mind, especially considering we're at a relatively early stage of this project.

Since so much development starts by copying a previously defined file and modifying it, I agree it's in our best interest to have the source as clean as possible.

.github/workflows/pr_develop.yml
```diff
@@ -85,6 +92,12 @@ jobs:
       - prepare
     uses: ./.github/workflows/check_lint.yml
 
+  check_format:
+    name: Formatting
+    needs:
+      - prepare
+    uses: ./.github/workflows/check_formatting.yml
+
```
gloopydoop, line 100: or maybe this is the crux hahaha.
either way, I'm on board

gloopydoop (commented):
You listed three points in this PR. The second two I think I understand, and I'm on board with. I don't understand what the first one is?
What is a "merge queue"?

actually you explain it pretty well in the description hahahaha.

I just couldn't find it on the code

gloopydoop:
Also I didn't click approve yet because I think for something like this it's best to have a unanimous decision, so let's wait for Abbas to review this too

.github/workflows/check_formatting.yml
timfelle: It will add some extra work in the beginning yes. But it is only applied to the files that have been modified. Additionally the same check is also in progress for making it into Neko, which helps us when extending functionality here as well.

timfelle: I added some text in the PR description now. It just makes sure multiple PR's accepted within a short time does not break each other

AbbasMoosaV (commented):
This looks good! The only issue is that I don't have sudo access on my system, so the dependency installation (e.g., findent) might fail for me. Maybe add a check to run sudo commands only if sudo access is available?

timfelle:
> This looks good! The only issue is that I don't have sudo access on my system, so the dependency installation (e.g., findent) might fail for me. Maybe add a check to run sudo commands only if sudo access is available?

This stuff is not done on your local machine. It is done on the github servers

AbbasMoosaV (approved):

PR #82, opened 2025-02-04
Fix MPI synchronization bug and clean up codebase

timfelle:
Address a bug that caused MPI de-synchronization and enhance code clarity through multiple cleanup rounds. Additionally, update example configurations and refine precision settings.

### DIFF
--- a/examples/rugby_verification/rugby_ball.case
+++ b/examples/rugby_verification/rugby_ball.case
@@ -4,7 +4,7 @@
         "mesh_file": "box.nmsh",
         "output_checkpoints": false,
         "output_at_end": false,
-        "output_boundary": true,
+        "output_boundary": false,
         "end_time": 0.1,
         "timestep": 1e-4,
         "numerics": {
@@ -23,12 +23,7 @@
                 {
                     "type": "no_slip",
                     "zone_indices": [ 1, 2 ]
-                },
-                // {
-                //     "type": "velocity_value",
-                //     "value": [ 0, 0, 0 ],
-                //     "zone_indices": [ 2 ]
-                // }
+                }
             ],
         },
         "fluid": {
@@ -99,7 +94,6 @@
             {
                 "type": "lube_term",
                 "weight": 1.0,
-                // "mask_name": "optimization_domain",
                 "K": 1.0
             }
         ],
@@ -113,7 +107,7 @@
         ],
     },
     "mma": {
-        "max_iter": 5,
+        "max_iter": 100,
         "asyinit": 0.5,
         "asyincr": 1.2,
         "asydecr": 0.7,
--- a/examples/rugby_verification/rugby_ball.case
+++ b/examples/rugby_verification/rugby_ball.case
@@ -4,7 +4,7 @@
         "mesh_file": "box.nmsh",
         "output_checkpoints": false,
         "output_at_end": false,
-        "output_boundary": false,
+        "output_boundary": true,
         "end_time": 0.1,
         "timestep": 1e-4,
         "numerics": {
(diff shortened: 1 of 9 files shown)


examples/rugby_verification/rugby_ball.case
```diff
@@ -81,8 +90,30 @@
...
+                "type": "volume",
+                "mask_name": "optimization_domain",
+                "limit": 0.2,
+                "is_max": false
+            }
+        ],
+    },
     "mma": {
-        "max_iter": 10,
+        "max_iter": 5,
```
AbbasMoosaV: I don't think these sort of changes to the case file should be in the PR.

sources/optimizer/bcknd/cpu/mma_cpu.f90
```diff
@@ -189,223 +178,220 @@ subroutine mma_subsolve_dpip_cpu(this, designx)
...
-       residunorm = sqrt(norm2(residu_small)**2 + re_xstuff_squ_global)
+       call MPI_Allreduce(MPI_IN_PLACE, re_sq_norm, &
+            1, mpi_real_precision, mpi_sum, neko_comm, ierr)
 
+       residual_norm = sqrt(norm2(residual_small)**2 + re_sq_norm)
+
+       ! --------------------------------------------------------------------- !
+       ! Internal loop
+
+       do iter = 1, this%max_iter
```
AbbasMoosaV, line 302: This%max_iter is the maximum number of iterations for the optimization loop in the optimizer%run() function.
Here for the internal loop in subsolv, I think we should set this to 100.
Maybe later on we can add another max_iter_internal_subsolv to json, but I don't think it is necessary.
In most cases, it is better if user does not play with these detail settings in the solver.

`do iter = 1, 100`

sources/optimizer/bcknd/cpu/mma_cpu.f90
```diff
@@ -189,223 +178,220 @@ subroutine mma_subsolve_dpip_cpu(this, designx)
...
+    mu = max(1.0_rp, 0.5_rp * this%c%x)
+
+    call MPI_Allreduce(this%n, nglobal, 1, &
+         MPI_INTEGER, mpi_sum, neko_comm, ierr)
+
+    ! ------------------------------------------------------------------------ !
+    ! Computing the minimal epsilon and choose the most conservative one
+
+    minimal_epsilon = max(0.9_rp * this%epsimin, 1.0e-12_rp)
+    call MPI_Allreduce(MPI_IN_PLACE, minimal_epsilon, 1, &
```
AbbasMoosaV: Is this helping with the MPI stuff? I don't see why we need this.
I think we should either set this%epsimin to a fixed number, e.g. 1.0e-9_rp, or we should use the total number of the design variables according to the formula:
`1.0e-9_rp * sqrt(real(m + nglobal, rp))`

sources/optimizer/bcknd/cpu/mma_cpu.f90
AbbasMoosaV, line 1: Overall, I really like the updates to this file, its much easier to read now.
Also based on my tests, we get a slightly better performance as well.

sources/optimizer/bcknd/cpu/mma_cpu.f90
timfelle: Sure i will do the global stuff

sources/optimizer/bcknd/cpu/mma_cpu.f90
timfelle: I will add a max_iter on the optimizer level to account for this.
This should be a variable we can work with, but as you mention it is an internal
parameter for MMA. Not a parameter for the optimizer.

AbbasMoosaV (approved):

PR #84, opened 2025-02-06
Fix buggy imports

timfelle:
Refactor module imports to ensure proper usage and avoid conflicts. This improves code clarity and maintains consistency across modules.

### DIFF
--- a/examples/mma/mma_simcomp.f90
+++ b/examples/mma/mma_simcomp.f90
@@ -32,7 +32,6 @@
 
  ! Implements the `mma_comp_t` type.
 module mma_simcomp
-  use neko_config
   use num_types, only: rp
   use case, only: case_t
   use json_module, only: json_file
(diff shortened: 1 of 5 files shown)


gloopydoop (approved):
ok all this looks reasonable.

Moving forward, 2 things:
1) so we should always have our imports at the top of the modules, correct?
2) I guess it's frowned upon to do `use module` compared to `use module, only: subroutine`
Are you checking this stuff manually or somehow automatically?

timfelle:
> ok all this looks reasonable.
> 
> Moving forward, 2 things:
> 
>     1. so we should always have our imports at the top of the modules, correct?
> 
>     2. I guess it's frowned upon to do `use module` compared to `use module, only: subroutine`
>        Are you checking this stuff manually or somehow automatically?

Yes, that is a better way to keep track of.
I am checking it manually yes. I think the linter technically looks for it, but our linter is not strict at all.

PR #86, opened 2025-02-07
Enhance Fortran build configuration and add LAPACK interface

timfelle:
Improve debugging options in the Fortran build configuration and introduce a new LAPACK interface module for solving linear equations. Additionally, refine type declarations for better clarity in the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,13 +83,28 @@ set(PFUNIT_DIR       ${CMAKE_CURRENT_SOURCE_DIR}/external/pFUnit
 set(CMAKE_Fortran_PREPROCESS ON)
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
     add_compile_options(
+        # Debugging flags
+        $<$<COMPILE_LANGUAGE:Fortran>:-g>
         $<$<COMPILE_LANGUAGE:Fortran>:-Og>
+
+        # Warnings which for debug builds are errors
         $<$<COMPILE_LANGUAGE:Fortran>:-Wall>
         $<$<COMPILE_LANGUAGE:Fortran>:-Wextra>
-        $<$<COMPILE_LANGUAGE:Fortran>:-Wno-unused-dummy-argument>
-        $<$<COMPILE_LANGUAGE:Fortran>:-Wno-function-elimination>
-        $<$<COMPILE_LANGUAGE:Fortran>:-Wno-missing-include-dirs>
         $<$<COMPILE_LANGUAGE:Fortran>:-Werror>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wsurprising>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wno-array-temporaries>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wimplicit-interface>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wno-unused-dummy-argument>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wfunction-elimination>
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -83,13 +83,28 @@ set(PFUNIT_DIR       ${CMAKE_CURRENT_SOURCE_DIR}/external/pFUnit
...
         $<$<COMPILE_LANGUAGE:Fortran>:-Wno-missing-include-dirs>
-        $<$<COMPILE_LANGUAGE:Fortran>:-Werror>
+        $<$<COMPILE_LANGUAGE:Fortran>:-Wcompare-reals>
+
+        # Debugging checks
+        $<$<COMPILE_LANGUAGE:Fortran>:-fcheck=all>
+        $<$<COMPILE_LANGUAGE:Fortran>:-fbacktrace>
+        $<$<COMPILE_LANGUAGE:Fortran>:-ffpe-trap=invalid,zero,overflow>
+        $<$<COMPILE_LANGUAGE:Fortran>:-finit-real=snan>
+        $<$<COMPILE_LANGUAGE:Fortran>:-finit-integer=-99999999>
```
gloopydoop, line 107: am I understanding this correctly, these are related to uninitialized variables?

sources/adjoint/adjoint_pnpn.f90
```diff
@@ -79,7 +79,7 @@ module adjoint_pnpn
   use file, only: file_t
   use operators, only: ortho
   use vector, only: vector_t
-  use device_math, only: device_vlsc3
+  use device_math, only: device_vlsc3, device_cmult
```
gloopydoop, line 82: :+1:

sources/neko_ext/develop.f90
```diff
@@ -70,7 +70,10 @@ pure function area(nv, vertices) result(a)
     v1 = vertices(:, 1)
     cp = 0.0
     do v = 2, nv - 1
-       cp = cp + 0.5 * cross((vertices(:, v) - v1), (vertices(:, v + 1) - v1))
+       e1 = (vertices(:, v) - v1)
+       e2 = (vertices(:, v + 1) - v1)
+       tmp = cross(e1, e2)
+       cp = cp + 0.5 * tmp
```
gloopydoop, line 76: sure, these are equivalent :+1:
Is there a reason for the switch? Legibility? or does this somehow improve performance?

gloopydoop (approved):
OK I see 3 (kinda 4) changes here.

1) debug compiler flags. These seems reasonable (although, this is very much your forte, so I trust you completely here)
2) A lot of changes from lengthy computations in one line, to breaking them apart with working arrays. I doubled checked them all, they all seem ok. Is the reason for this legibility or efficiency?
3) The Lapack interface. This seems great, I guess you're moving in the direction of solving arbitrary optimization problems? This looks great

(sort of 4) The last device inclusion. I can confirm that this (along with PR#87) compiles fine on CUDA.

Good job :+1:

sources/neko_ext/develop.f90
timfelle: This is to start avoiding temporaries arrays defined implicitly by the compiler. It should help us monitor the memory footprint etc more easily. It should improve performance as the compiler have an easier time knowing the array sizes at compile time instead

CMakeLists.txt
timfelle: They are related to a lot of different errors. Overflowing variables, forcing a full backtrace to allow easier debugging on errors and such

gloopydoop: :+1:

sources/neko_ext/develop.f90
gloopydoop: :+1:

PR #87, opened 2025-02-07
Fix CMake compile definition and preprocessor directives for CUDA

timfelle:
Correct the application of the compile definition for CUDA in CMake and update preprocessor directives to ensure proper conditional compilation.


gloopydoop (approved):
:star: superstar!!! :star:

PR #89, opened 2025-02-10
Revert Fortran build configuration changes and improve debugging options

timfelle:
Revert previous enhancements to the Fortran build configuration while relaxing debug options and addressing identified issues with array temporaries.


CMakeLists.txt
```diff
@@ -80,6 +80,13 @@ set(PFUNIT_DIR       ${CMAKE_CURRENT_SOURCE_DIR}/external/pFUnit
 # ............................................................................ #
 # Define compiler flags
 
+# Define the list of available build types
+set(CMAKE_BUILD_TYPE_LIST
+    Debug
+    Release
+    Testing
+)
+
```
gloopydoop, line 89: Is this the fix for `CMAKE_BUILD_TYPE=Debug ./setup.sh`?

gloopydoop: ok so for clarity, release is the least strict, I know that.

out of debug and testing, which one does github check?
out of release and testing, which one should we be compiling with to ensure we're not introducing major bugs etc.

sources/optimizer/bcknd/cpu/mma_cpu.f90
```diff
@@ -251,11 +249,8 @@ subroutine mma_subsolve_dpip_cpu(this, designx)
             a0 => this%a0, a => this%a%x, &
             bi => this%bi%x)
 
-         work_1 = matmul(transpose(pij), lambda)
-         work_2 = matmul(transpose(qij), lambda)
-
-         rex = (p0j + work_1) / (upp - x)**2 &
-              - (q0j + work_2) / (x - low)**2 &
+         rex = (p0j + matmul(transpose(pij), lambda)) / (upp - x)**2 &
+              - (q0j + matmul(transpose(qij), lambda)) / (x - low)**2 &
```
gloopydoop, line 253: now I'm super confused hahahaha.
This is undo'ing a previous PR I reviewed :sweat_smile:

What was the problem here?

gloopydoop (approved):

CMakeLists.txt
timfelle: This is not the fix.
Moving the stricter -W and -f flags to Testing fixes the issue.

Github will compile with the `Debug` variant.
You should run with `Debug` when you are debugging your code, and you should run with `Release` when you are producing results and testing outputs of the code.

The `Testing` variant is for extra strict checking and will be used for some performance optimization, as it checks for invalid array indexing and temporary allocation of variables.

sources/optimizer/bcknd/cpu/mma_cpu.f90
timfelle: The way we defined the `work_1` and `work_2` arrays caused some corruption of the underlying list definition, resulting in a segmentation fault.

gloopydoop: hmmm interesting, ok let's talk in more detail on Tuesday

PR #91, opened 2025-02-11
Improved logging

gloopydoop:
I guess we discussed this a bit this morning, but this is what I had in mind.
I'm putting this in draft mode because I want comments before I move any further with this:

- Do we think we'll want this in other optimization? ie, would a `logger_t` be useful?
- Are we all happy with enforcing a log name to all constraints/objectives so we can write them out?

But perhaps more importantly, what I've written currently is pretty susceptible to truncation of that long string (buffer? is this the correct word? that was the intention of the variable name `buff`)

### Example output

```
n =     86400, m =         1, tolerance =  1.00000000000000002E-03              
iter, Total objective function, Dissipation, out of plane stresses, Volume constraint, KKTmax, KKTnorm2, scalaing factor
  0,     5.3696209634E+01,     4.0441625841E+01,     1.3254583793E+01,    -3.4113716238E-01,     1.7976931349+308,     1.7976931349+308,     1.0000000000E+00
  1,     5.3631210218E+01,     4.0441625841E+01,     1.3189584378E+01,    -3.4067827356E-01,     3.4063388244E-01,     1.8251180684E+00,     1.0000000000E+00
  2,     5.3266158113E+01,     4.0127814834E+01,     1.3138343279E+01,    -3.4021952588E-01,     3.4017508569E-01,     5.0791113981E+00,     1.0000000000E+00
  3,     5.2946050791E+01,     3.9894493099E+01,     1.3051557691E+01,    -3.3973026153E-01,     3.3968611728E-01,     4.1537413064E+00,     1.0000000000E+00
  4,     5.2488335978E+01,     3.9561169448E+01,     1.2927166530E+01,    -3.3922171690E-01,     3.3917795103E-01,     5.6123647507E+00,     1.0000000000E+00
  5,     5.1825567047E+01,     3.9082081599E+01,     1.2743485448E+01,    -3.3889046816E-01,     3.3884813432E-01,     7.9908112190E+00,     1.0000000000E+00
  6,     5.1000305086E+01,     3.8491289553E+01,     1.2509015532E+01,    -3.3852679847E-01,     3.3848677945E-01,     9.8497959414E+00,     1.0000000000E+00
  7,     5.0354702281E+01,     3.8025030555E+01,     1.2329671726E+01,    -3.3866646158E-01,     3.3862891930E-01,     7.8845237703E+00,     1.0000000000E+00
  8,     4.9785292383E+01,     3.7615237248E+01,     1.2170055135E+01,    -3.3902719413E-01,     3.3899308096E-01,     7.5623243319E+00,     1.0000000000E+00

```

### Example output using `csv_file_type`
```
iter, Total objective function, Dissipation, Out of plane stresses, Volume constraint, KKTmax, KKTnorm2, scalaing factor
0.0000000000000000,53.696209634065141,40.441625840817053,13.254583793248088,-0.34113716238378983,0.0000000000000000,0.0000000000000000,1.0000000000000000
1.0000000000000000,53.631210218451116,40.441625840817053,13.189584377634059,-0.34067827356064134,0.34063388243730330,1.8251180684119177,1.0000000000000000
2.0000000000000000,53.266158113095081,40.127814833617478,13.138343279477603,-0.34021952587850796,0.34017508568651300,5.0791113980976883,1.0000000000000000
3.0000000000000000,52.946050790610592,39.894493099386999,13.051557691223593,-0.33973026152689706,0.33968611727815884,4.1537413063900503,1.0000000000000000
4.0000000000000000,52.488335977948438,39.561169447724346,12.927166530224092,-0.33922171690237829,0.33917795102621462,5.6123647507375267,1.0000000000000000
5.0000000000000000,51.825567047220396,39.082081599088639,12.743485448131759,-0.33889046815908813,0.33884813432446081,7.9908112190203324,1.0000000000000000
6.0000000000000000,51.000305085635944,38.491289553207920,12.509015532428020,-0.33852679846645700,0.33848677945129146,9.8497959414055956,1.0000000000000000
7.0000000000000000,50.354702281281916,38.025030555367294,12.329671725914622,-0.33866646158438390,0.33862891930457578,7.8845237703490136,1.0000000000000000
8.0000000000000000,49.785292383215989,37.615237248149093,12.170055135066898,-0.33902719412641674,0.33899308096021136,7.5623243318792319,1.0000000000000000
```

### DIFF
--- a/sources/optimizer/mma_optimizer.f90
+++ b/sources/optimizer/mma_optimizer.f90
@@ -147,7 +147,7 @@ contains
     call problem%get_all_objective_values(all_objectives)
 
     ! Stamp the fist iteration
-    call mma_logger_assemble_data(log_data, 1, objective_value, &
+    call mma_logger_assemble_data(log_data, 0, objective_value, &
          all_objectives, constraint_value, 0.0_rp, 0.0_rp, scalingfactor, &
          problem%get_n_objectives(), problem%get_n_constraints())
     call this%logger%write(log_data)
@@ -156,7 +156,7 @@ contains
 
     associate(x => this%design%design_indicator%x)
 
-      do iter = 2, max_iter
+      do iter = 1, max_iter
          if (this%mma%get_residumax() .lt. tolerance) exit
 
          !Scaling
--- a/sources/optimizer/mma_optimizer.f90
+++ b/sources/optimizer/mma_optimizer.f90
@@ -24,7 +24,7 @@ module mma_optimizer
   use neko_ext, only: reset
   use mask_ops, only: mask_exterior_const
 
-  use comm, only : pe_rank
+  use csv_file, only : csv_file_t
 
 
   implicit none
@@ -36,6 +36,7 @@ module mma_optimizer
 
      type(mma_t) :: mma
      type(topopt_design_t), pointer :: design
+     type(csv_file_t) :: logger
 
      !> Scaling constraint_value%x and constraint_sensitivities%x.
      !! Note that the values are not updated but they are scaled when passed
@@ -112,25 +113,21 @@ contains
     real(kind=rp) :: scalingfactor
 
     real(kind=rp) :: objective_value
+    type(vector_t) :: all_objectives
     type(vector_t) :: constraint_value
     type(vector_t) :: objective_sensitivities
     type(matrix_t) :: constraint_sensitivities
 
-    ! These would all be owned by the logger_t
-    ! ------------------------------------------------------------------------
-    !> format for writing real values in the log file
-    character(len=10) :: real_format
-    ! logging
-    character(len=1024) :: problem_log ! I assume 1024 is enough?
-    character(len=1024) :: optimization_log ! I assume 1024 is enough?
-    character(len=100) :: log_format
-    integer :: log_unit
+    character(len=1024) :: optimization_header
+    character(len=1024) :: problem_header
+    type(vector_t) :: log_data
     ! ------------------------------------------------------------------------
 
 
     ! call MPI_Comm_rank(neko_comm, rank, ierr)
     call MPI_Allreduce(this%mma%get_n(), nglobal, 1, &
          MPI_INTEGER, mpi_sum, neko_comm, ierr)
+
 
     !>initializing the scaling factor
     scalingfactor = 1.0_rp
@@ -143,46 +140,27 @@ contains
     call problem%get_constraint_values(constraint_value)
     call problem%get_objective_sensitivities(objective_sensitivities)
     call problem%get_constraint_sensitivities(constraint_sensitivities)
+    call problem%get_all_objective_values(all_objectives)
 
-    ! Should I write a logger_t such that this becomes a logger%init()
-    ! ------------------------------------------------------------------------
-    !Writing the optimization data in a separate file
-    log_unit = 1368
-    real_format = 'ES20.10'
-    open(log_unit, file = "optimization_data.txt", status = "replace")
-    log_format = '(I3, A, ", ", '//real_format//', ", ", '//real_format//&
-       ', ", ", '//real_format//')'
-    ! Write n, m, and tolerance in the first line of optimization_data.txt
-    ! do we really want this?
-    if (pe_rank .eq. 0) then
-       write(log_unit, '("n =", I10, ", m =", I10, ", tolerance =", ES25.17)') &
-            nglobal, this%mma%get_m(), tolerance
-    end if
 
-    ! Write the header for the remaining data
-    problem_log = problem%get_log_header()
-    optimization_log = 'iter, '//trim(problem_log)//&
+    ! Initialize the logger
+    call this%logger%init('optimization_data.txt')
+
+    ! Write the header
+    problem_header = problem%get_log_header()
+    optimization_header = 'iter, '//trim(problem_header)//&
        ', KKTmax, KKTnorm2, scalaing factor'
-    if (pe_rank .eq. 0) then
-       write(log_unit, '(A)') trim(optimization_log)
-    end if
-    ! ------------------------------------------------------------------------
+    call this%logger%set_header(trim(optimization_header))
+
+    ! Stamp the zeroth iteration
+    call mma_logger_assemble_data(log_data, 0, objective_value, &
+       all_objectives, constraint_value, 0.0_rp, 0.0_rp, scalingfactor, &
+       problem%get_n_objectives(), problem%get_n_constraints())
+    call this%logger%write(log_data)
+
+    call problem%write(0)
 
     associate(x => this%design%design_indicator%x)
-
-
-      ! Should I write a logger_t such that this becomes a logger%write()
-      ! ----------------------------------------------------------------------
-      ! Write the data row-by-row
-      problem_log = problem%get_log_state(real_format)
-      write(optimization_log, log_format) 0, trim(problem_log), &
-         this%mma%get_residumax(), this%mma%get_residunorm(), scalingfactor
-      if (pe_rank .eq. 0) then
-         write(1368, '(A)') trim(optimization_log)
-      end if
-      ! ----------------------------------------------------------------------
-
-      call problem%write(0)
 
       do iter = 1, max_iter
          if (this%mma%get_residumax() .lt. tolerance) exit
@@ -211,23 +189,18 @@ contains
          call problem%get_constraint_values(constraint_value)
          call problem%get_objective_sensitivities(objective_sensitivities)
          call problem%get_constraint_sensitivities(constraint_sensitivities)
+         call problem%get_all_objective_values(all_objectives)
 
          call this%mma%KKT(x, objective_sensitivities%x, &
               reshape([constraint_value%x], [this%mma%get_m()]), &
               constraint_sensitivities%x)
 
-         ! Write the data row-by-row
-         ! should I implement a logger_t, such that this is a logger%write()
-         ! -------------------------------------------------------------------
-         problem_log = problem%get_log_state(real_format)
-         write(optimization_log, log_format) iter, trim(problem_log), &
-            this%mma%get_residumax(), this%mma%get_residunorm(), scalingfactor
-         if (pe_rank .eq. 0) then
-            write(log_unit, '(A)') trim(optimization_log)
-            ! Flush the buffer to write the data during the run
-            flush(log_unit)
-         end if
-         ! -------------------------------------------------------------------
+         ! Stamp the zeroth iteration
+         call mma_logger_assemble_data(log_data, iter, objective_value, &
+            all_objectives, constraint_value, this%mma%get_residumax(), &
+            this%mma%get_residunorm(), scalingfactor, &
+            problem%get_n_objectives(), problem%get_n_constraints())
+         call this%logger%write(log_data)
 
          call problem%write(iter)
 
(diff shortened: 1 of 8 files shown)


timfelle (commented):
> I guess we discussed this a bit this morning, but this is what I had in mind.
> I'm putting this in draft mode because I want comments before I move any further with this:
> 
> - Do we think we'll want this in other optimization? ie, would a `logger_t` be useful?
> - Are we all happy with enforcing a log name to all constraints/objectives so we can write them out?
> 
> But perhaps more importantly, what I've written currently is pretty susceptible to truncation of that long string (buffer? is this the correct word? that was the intention of the variable name `buff`) 

In general I think we should use the logger type defined by neko. Not the neko_log object, but the log_t. Since that one contains all the things we would want in terms of logging.
To save the history of optimization and values for objectives and all that, I would say we should use the `csv` file writer already defined in Neko.
For now this is a nice little improvement.

sources/problems/base_functional.f90
```diff
@@ -57,7 +57,8 @@ module base_functional
      real(kind=rp) :: value
      !> Sensitivity field
      type(vector_t) :: sensitivity
-
+     !> Name of constraint/objective in the logfile
+     character(len=25) :: log_name
```
timfelle: Why not just call it the `name` of the functional.

gloopydoop: I thought about that, but then I figured there could be a circumstance that you would have double ups of the same thing. eg, a minimum volume constraint and a maximum volume constraint. Or two of the same objective but defined in different regions etc.

sources/problems/problem.f90
```diff
@@ -68,6 +68,7 @@ module problem
      integer :: n_objectives
      !> Number of constraints in the problem.
      integer :: n_constraints
+     !> Format for output log
```
timfelle: What are you documenting here?
`!>` is the command for doxygen to add the comment as the documentation of the
object that follow. Meaning the `objective_list` now gets this added to it's
documentation, which i doubt is your intention.

sources/problems/problem.f90
```diff
@@ -595,4 +601,77 @@ pure function problem_get_num_constraints(this) result(n)
...
+            this%constraint_list(i)%constraint%log_name
+       buff = trim(buff)//trim(mini_buff)
+    end do
+
+  end function problem_get_log_header
+
+  !> Return the log for the current iteration.
+  function problem_get_log_state(this, real_format) result(buff)
+    class(problem_t), intent(in) :: this
+    character(len=*), intent(in) :: real_format
```
timfelle: Why not just call it `format`, what makes it `real` and why is that different
from the other formats you supply as inputs?

sources/problems/problem.f90
gloopydoop: oh snap you're right, yeh that should be deleted. I had originally tried giving the format to the problem, but then I figured that was more something that should belong to the optimizer and be passed into the problem.

sources/problems/objective.f90
```diff
@@ -89,15 +89,19 @@ end subroutine objective_factory
   !! @param design_size The number of design variables.
   !! @param weight The weight of the objective function.
   !! @param[optional] mask_name The name design the mask.
-  subroutine objective_init_base(this, design_size, weight, mask_name)
+  subroutine objective_init_base(this, design_size, weight, log_name, &
```
timfelle: Why is the name of the log added as an input to the initializer?

sources/optimizer/mma_optimizer.f90
```diff
@@ -24,6 +24,8 @@ module mma_optimizer
   use neko_ext, only: reset
   use mask_ops, only: mask_exterior_const
 
+  use comm, only : pe_rank
```
timfelle: Why do you add a new line grabbing stuff from `comm` when we already have a line
doing so further up?

sources/optimizer/mma_optimizer.f90
```diff
@@ -183,12 +216,18 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance, &
...
               constraint_sensitivities%x)
 
-         write(1368, '(I3, ",", ES25.17, ",", ES25.17, ",", ES25.17, ",", &
-              & ES25.17, ",", ES25.17)') iter, objective_value, constraint_value%x, &
-              this%mma%get_residumax(), this%mma%get_residunorm(), scalingfactor
-
-         ! Flush the buffer to write the data during the run
-         flush(1368)
+         ! Write the data row-by-row
+         ! should I implement a logger_t, such that this is a logger%write()
```
timfelle: No. Use the csv_t from neko

gloopydoop: working on a new PR, hold up fam :+1:

sources/problems/constraints/volume_constraint.f90
```diff
@@ -122,14 +122,17 @@ subroutine volume_constraint_init_json(this, json, design, simulation)
     type(simulation_t), target, intent(inout) :: simulation
 
     character(len=:), allocatable :: mask_name
+    character(len=:), allocatable :: log_name
     logical :: is_max
     real(kind=rp) :: limit
 
     call json_get_or_default(json, "mask_name", mask_name, "")
+    call json_get_or_default(json, "log_name", log_name, "Volume constraint")
```
timfelle: This should not be set by the user. A volume constraint is a volume constraint.

sources/problems/base_functional.f90
timfelle: When i read the variable name `log_name`, i expect that this is the name of some
log. Not that it is the name of the current instance of the object, which
happens to be used in a log.

You can easily have different names for different objects if the variable is
called `name`

sources/optimizer/mma_optimizer.f90
```diff
@@ -34,6 +36,7 @@ module mma_optimizer
 
      type(mma_t) :: mma
      type(topopt_design_t), pointer :: design
+     type(csv_file_t) :: logger
```
timfelle: Please put this on the top-level optimizer instead of the MMA_optimizer. it is
not special to the MMA one.

sources/optimizer/mma_optimizer.f90
```diff
@@ -110,10 +112,15 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance, &
...
 
     real(kind=rp) :: objective_value
+    type(vector_t) :: all_objectives
     type(vector_t) :: constraint_value
     type(vector_t) :: objective_sensitivities
     type(matrix_t) :: constraint_sensitivities
 
+    character(len=1024) :: optimization_header
+    character(len=1024) :: problem_header
+    type(vector_t) :: log_data
```
timfelle: The headers should be handled by the initializer, not the runner.

sources/optimizer/mma_optimizer.f90
```diff
@@ -129,27 +136,26 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance, &
...
+    ! Initialize the logger
+    call this%logger%init('optimization_data.txt')
 
-      ! Write n, m, and tolerance in the first line of optimization_data.txt
-      write(1368, '("n =", I10, ", m =", I10, ", tolerance =", ES25.17)') &
-           nglobal, this%mma%get_m(), tolerance
+    ! Write the header
+    problem_header = problem%get_log_header()
+    optimization_header = 'iter, '//trim(problem_header)//&
+       ', KKTmax, KKTnorm2, scalaing factor'
```
timfelle: Please use a csv file, not a txt file.

sources/problems/base_functional.f90
gloopydoop: ah I thought you meant "the `name` of the functional" you meant the `volume_constraint` or `lube_term_objective` etc.

`name` is good :+1:

sources/optimizer/mma_optimizer.f90
```diff
@@ -178,17 +184,18 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance, &
...
-              this%mma%get_residumax(), this%mma%get_residunorm(), scalingfactor
-
-         ! Flush the buffer to write the data during the run
-         flush(1368)
+         ! Stamp the i^th iteration
+         call mma_logger_assemble_data(log_data, iter, objective_value, &
+            all_objectives, constraint_value, this%mma%get_residumax(), &
+            this%mma%get_residunorm(), scalingfactor, &
+            problem%get_n_objectives(), problem%get_n_constraints())
+         call this%logger%write(log_data)
```
timfelle, line 197: This is not the zeroth iteration

sources/optimizer/mma_optimizer.f90
gloopydoop: yeh this is a good point.
I've put it in the top level optimizer, but it's public, which I think is ok...

I half suspect you would have preferred to keep the logger private and have some interface around it?

sources/problems/objective.f90
gloopydoop: yeh this one in the init_base I've gotten rid of

sources/problems/constraints/volume_constraint.f90
gloopydoop: This I disagree. if you had $$V_{min} < \int_\Omega \rho < V_{max}$$ you would split into two constraints, both of the same type, but one would be a minimum volume and one would be a maximum volume. In that case it would be nice to name them differently in the logfile.

Similar edge cases could come up with the masks

sources/optimizer/mma_optimizer.f90
timfelle: Yes, but we can get back to that later.

sources/optimizer/mma_optimizer.f90
```diff
@@ -129,29 +144,19 @@ subroutine mma_optimizer_run_steady_state_prob(this, problem, tolerance, &
...
-      write(1368, '(I3, ",", ES25.17, ",", ES25.17, ",", ES25.17, ",", &
-           & ES25.17, ",", ES25.17)') &
-           0, objective_value, constraint_value%x, &
-           this%mma%get_residumax(), this%mma%get_residunorm(), scalingfactor
-
-      call problem%write(0)
+    associate(x => this%design%design_indicator%x)
 
-      do iter = 1, max_iter
+      do iter = 2, max_iter
```
timfelle: Why do you start at iteration 2?

gloopydoop: Yeh this is a good question, and we should all talk about. let's get all three of us involved since @AbbasMoosaV wrote the initial loop, and maybe he interpreted things differently, but the overall structure of the loop is strange and should be revisited in a different PR.

In the previous commit I had this starting from 1, and stamped the "zeroth" iteration.
I was calling the "zeroth" iteration just running forward, no MMA.

currently this loop goes:
run forward
enter the loop
   run mma
   run forward
   stamp
   reset simulation


I would probably argue a restructuring to:
enter the loop
   run forward
   run mma
   stamp
   reset simulation

would be more appropriate. But I may have missed something so I'll wait for Abbas.
Anyway, it's unrelated to this PR so I'm going to switch back to iter going from 1, max_iter and stamping the "zeroth" iteration, and add that restructuring to a new PR.

timfelle: Just call it Initial conditions instead of "zeroth" iteration.

timfelle (approved):

PR #92, opened 2025-02-13
Allow specification of processor count for execution

timfelle:
Introduce the ability to specify the number of processors to use via the `NPROCS` environment variable, enhancing flexibility in resource allocation during script execution.


gloopydoop (commented):
Just by the description I know I'm going to like this PR :sweat_smile: 

Question before I look more closely, currently I use `run.sh` to do this. Does this in some way override `run.sh` or will `run.sh` override this?

**edit**

Sorry that probably wasn't super clear, I mean a `run.sh` that lives in the example

timfelle:
The one in the example always take precedence. Since if that is there, you clearly wish to have granular control

gloopydoop (approved):
yehaw :cowboy_hat_face:

PR #93, opened 2025-02-13
Refactor/design-problem

timfelle:
This PR build on the structure.

The design, problem and simulation is untangled. The steady state problem are no longer needed.


sources/optimizer/mma_optimizer.f90
```diff
@@ -288,12 +288,13 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
...
 
-       call problem%write(iter)
+       call simulation%write(iter)
+       call design%write(iter)
        call simulation%reset()
     end do
 
     ! Final state after optimization
-    print*, "MMA Optimization completed after", iter-1, "iterations."
+    print*, "MMA Optimization completed after", iter - 1, "iterations."
```
```suggestion
    print*, "MMA Optimization completed after", iter-1, "iterations."
```

sources/designs/design.f90
```diff
@@ -63,6 +75,23 @@ subroutine design_init_from_json(this, parameters, simulation)
...
+       import design_t, vector_t
+       class(design_t), intent(inout) :: this
+       type(vector_t), intent(in) :: sensitivity
+     end subroutine design_map_backward
+
+     subroutine design_write(this, idx)
+       import design_t
+       class(design_t), intent(inout) :: this
+       integer, intent(in) :: idx
+     end subroutine design_write
```
gloopydoop, line 94: Ok I see one problem with this, what exactly are we calling a `design` compared to a `topopt_design`? And how do we intend to separate an optimization problem involving a simulation and those which don't.

For some abstract optimization problem, the notion of mapping forward and backward might not make sense, and yet the `mma_optimizer` still requires a design. I kind of view this mapping as something specific to a topology optimization problem, so I feel it makes more sense having it part of the derived type.

thoughts?

sources/designs/topopt_design.f90
```diff
@@ -316,6 +320,33 @@ subroutine topopt_design_init_from_components(this, simulation)
...
+         simulation%adjoint_case%scheme%f_adj_y, &
+         simulation%adjoint_case%scheme%f_adj_z, &
+         this%brinkman_amplitude, &
+         simulation%adjoint_case%scheme%u_adj, &
+         simulation%adjoint_case%scheme%v_adj, &
+         simulation%adjoint_case%scheme%w_adj, &
+         simulation%adjoint_case%scheme%c_Xh)
+    ! append brinkman source term based on design
+    call simulation%adjoint_case%scheme%source_term%add(adjoint_brinkman)
+
```
gloopydoop, line 349: ok this is nice, so I guess this is what you meant by removing the steady_state_problem, it's the designs responsibility to insert himself into the simulation? correct?

Thinking towards the future and the passive scalar, I guess we should ultimately wrap these (or more likely the passive scalar equivalents) in `if` statements, first checking what kinds of schemes etc are enabled by the simulation, then inserting oneself. Again, job for the passive scalar PR, so you mark this resolved instantly.

sources/designs/topopt_design.f90
```diff
@@ -357,21 +387,26 @@ subroutine topopt_design_map_forward(this)
...
+    call neko_scratch_registry%request_field(df_dchi, temp_indices(1))
+    call copy(df_dchi%x, sensitivity%x, this%size())
 
     ! TODO
     ! again..
     ! so this would be:
     ! call mapper%backward(fld_out, fld_in)
     call neko_scratch_registry%request_field(dF_dfiltered_design, &
-         temp_indices(1))
+         temp_indices(2))
```
gloopydoop, line 409: yeh nice work with the registry. 

I guess one thing to be aware of in the map forward/backwards, notice that `%apply_backwards` requires not just the sensitivity coming in, but also the "unmapped" (I was calling it `X_in`, but I guess we've already discussed this is awkward naming convention) field.

So if you have a cascade of multiple mappings
$A \mapsto B \mapsto C \mapsto D$
when you chain rule yourself backwards you need to have C, B and A stored.

Again, this is for a future PR when we write these mappings as a "cascade" but we will likely end up needing a `field_list_t` the size of `n-1` where `n` is the number of operations in the cascade.

sources/designs/topopt_design.f90
```diff
@@ -405,12 +440,12 @@ subroutine topopt_design_free(this)
 
   end subroutine topopt_design_free
 
-  subroutine topopt_design_sample(this,t)
-    class(topopt_design_t), target, intent(inout) :: this
-    real(kind=rp), intent(in) :: t
+  subroutine topopt_design_write(this, idx)
+    class(topopt_design_t), intent(inout) :: this
+    integer, intent(in) :: idx
```
gloopydoop, line 445: this is very nice

sources/simulation/simulation.f90
```diff
@@ -102,6 +109,26 @@ subroutine simulation_init(this, parameters)
...
+    call this%output%init(sp, 'optimization', 8)
+    call this%output%fields%assign(1, this%fluid_scheme%p)
+    call this%output%fields%assign(2, this%fluid_scheme%u)
+    call this%output%fields%assign(3, this%fluid_scheme%v)
+    call this%output%fields%assign(4, this%fluid_scheme%w)
+    call this%output%fields%assign(5, this%adjoint_case%scheme%p_adj)
+    call this%output%fields%assign(6, this%adjoint_case%scheme%u_adj)
+    call this%output%fields%assign(7, this%adjoint_case%scheme%v_adj)
+    call this%output%fields%assign(8, this%adjoint_case%scheme%w_adj)
+
```
gloopydoop, line 135: have we lost functionality to visualize the the design now? Or the gradients?
I know it was a mess in `steady_state_problem`, and I can see why you want to keep the fields outputted by simulation to be exclusively related to simulation, but from a user/postprocessing perspective it's nice to have them all together.

gloopydoop (commented):
I've left some comments, mostly questions, but overall I think this is a nice untangling.

This time I haven't forgotten to press the approve button, I just haven't had time to checkout this branch and run some tests on the functionality. I'll do that after fika :+1:

sources/designs/design.f90
timfelle: The current design is explicitly a brinkman penalization based IBM for flow only problems.
Remember I have not made it to fix the design type. This PR is specifically for cleanup of the problem type.

We would need a new design type if we want to include, different conductivity for the design region. We would need a new design if we want to include convection terms.

sources/simulation/simulation.f90
timfelle: The idea is that the design will be in charge of writing itself. There are absolutely nothing requiring us to have a field to represent the design. We might actually want to not have that. Also Neko includes the "field_writer" simulation component, which can be used to write any field from the registry. So that would also be a way to output stuff like the brinkman amplitude field during a simulation.

sources/designs/topopt_design.f90
timfelle: Yes that is something that will require a bit of work to figure out properly.

sources/simulation/simulation.f90
```diff
@@ -102,6 +109,26 @@ subroutine simulation_init(this, parameters)
...
+    !---------------------------------------------------------
+    ! TODO
+    ! obviously when we do the mappings properly, to many coefficients, we'll
+    ! also have to modify this
+    ! for now:
+    ! - forward (p,u,v,w)                      1,2,3,4           p,vx,vy,vz
+    ! - adjoint (p,u,v,w)                      5,6,7,8           t,s1,s2,s3
+
+    ! Allocate the output type
+    call this%output%init(sp, 'optimization', 8)
```
gloopydoop: This should now no longer be called optimization.
Open to suggestions on this name, but I think `optimization` doesn't fit anymore.

I actually think, since we're splitting everything up anyway, having two field writers, one for forward, one for adjoint actually makes the most sense. Especially since visit etc let's you plot velocity fields with vectors more conveniently if they're in vx,vy,vz. I'll pop in an extra commit. (feel free to tweak it)

sources/simulation/simulation.f90
gloopydoop: I finally got the chance to pull and run this, yeh, I think it's good. We get the design and the simulation fields separately, they're synced correctly in terms of optimization iteration, so I think we're good to go.

The internal `neko` field writer is still bugging me (job for a future PR) ideally we would only have fields for forward, adjoint, design.

gloopydoop (commented):
I added one suggestion in a commit, but I think this is ready to be approved, I'll just wait to see what you think of the suggestion.

gloopydoop (approved):
I'm clicking the approve button but I disabled auto-merge, so hopefully this means you can have one last double check before anything goes through :+1:

PR #94, opened 2025-02-14
Fix verification run script and build examples from runner script

timfelle:
Streamline the execution process by building example executables from the runner script.

Update the verification example. New filenames were not used properly. Also the example will now check against the result folder, rather than a manually added reference.

Also added an MPI guard on a few print statements

### DIFF
--- a/sources/optimizer/mma_optimizer.f90
+++ b/sources/optimizer/mma_optimizer.f90
@@ -15,7 +15,7 @@ module mma_optimizer
   use matrix, only: matrix_t
 
   !only to print nglobal when running in parallel
-  use comm, only: neko_comm
+  use comm, only: neko_comm, pe_rank
   use mpi_f08, only: MPI_INTEGER, mpi_sum, MPI_Allreduce
 
   use neko_config, only: NEKO_BCKND_DEVICE
@@ -98,7 +98,9 @@ contains
        call neko_error('Unknown design type for MMA Optimizer')
     end select
 
-    print *, "Initializing mma_optimizer with steady_state_problem_t."
+    if (pe_rank .eq. 0) then
+       print *, "Initializing mma_optimizer with steady_state_problem_t."
+    end if
 
     call json_get_subdict(parameters, "optimization.solver", solver_parameters)
     call this%mma%init_json(x%x, &
@@ -160,7 +162,10 @@ contains
 
     !>initializing the scaling factor
     scaling_factor = 1.0_rp
-    print *, "max_iterations for the optimization loop = ", this%max_iterations
+    if (pe_rank .eq. 0) then
+       print *, "max_iterations for the optimization loop = ", &
+            this%max_iterations
+    end if
 
     call simulation%run_forward()
     call problem%compute(design)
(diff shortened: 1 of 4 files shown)


sources/optimizer/mma_optimizer.f90
```diff
@@ -98,7 +98,9 @@ subroutine mma_optimizer_init_from_json(this, parameters, problem, design, &
        call neko_error('Unknown design type for MMA Optimizer')
     end select
 
-    print *, "Initializing mma_optimizer with steady_state_problem_t."
+    if (pe_rank .eq. 0) then
+       print *, "Initializing mma_optimizer with steady_state_problem_t."
+    end if
```
gloopydoop, line 104: nice

gloopydoop: I guess we've lost a bit of consistency with the logging. We have 3 ways to log:

- the .csv writer (ends up in `optimization_data.csv`)
- the in built `neko` log (ends up in `rugby_ball.log`)
- these kind of print statements (ends up in `output.log`)

Is this the intended behavior for these print statements? or Just an inconsistency between how we've been logging?

gloopydoop (commented):
This is nice. 
One question: When testing I wrote:
`./run.sh rugby_verification`

And saw 

```
Compiling the examples.
[ 65%] Built target neko-top
[ 70%] Built target Examples-thermal_mix_channel
[ 74%] Built target Examples-control_flow
[ 79%] Built target Examples-rugby_verification
[ 81%] Built target Examples-cylinder_benchmark
[ 81%] Built target Examples-debug
[ 83%] Built target Examples-rugby_ball
[ 87%] Built target Examples-easy-E
[ 92%] Built target Examples-mma
[ 95%] Built target Examples-brinkman_term
[ 98%] Built target Examples-adjoint_example
[100%] Build all the examples.
[100%] Built target Examples

Queueing case files.
	Queued:      rugby_verification
	Started:     rugby_verification

```

Why are things like easy-E being built?

sources/optimizer/mma_optimizer.f90
timfelle: The `.csv` stuff is used to keep track of data which might be useful to the user afterwards. In this case, the optimization history.
The log that Neko creates is the log of Neko, so that is not the log of the optimization.
The print statements should be removed in time and placed in a separate log, similar to the Neko log. However, I have not had time to actually do that yet.

timfelle:
> This is nice. One question: When testing I wrote: `./run.sh rugby_verification`
> 
> And saw
> 
> ```
> Compiling the examples.
> [ 65%] Built target neko-top
> [ 70%] Built target Examples-thermal_mix_channel
> [ 74%] Built target Examples-control_flow
> [ 79%] Built target Examples-rugby_verification
> [ 81%] Built target Examples-cylinder_benchmark
> [ 81%] Built target Examples-debug
> [ 83%] Built target Examples-rugby_ball
> [ 87%] Built target Examples-easy-E
> [ 92%] Built target Examples-mma
> [ 95%] Built target Examples-brinkman_term
> [ 98%] Built target Examples-adjoint_example
> [100%] Build all the examples.
> [100%] Built target Examples
> 
> Queueing case files.
> 	Queued:      rugby_verification
> 	Started:     rugby_verification
> ```
> 
> Why are things like easy-E being built?

Because the examples are being compiled when you execute the run script. Because I thought it was a good idea to ensure the examples are up to date before we run an example. This will make sure you do not forget to compile the examples after making changes to the code.

gloopydoop (approved):
I keep forgetting the approve button :sweat_smile:

PR #103, opened 2025-02-24
Refactor the design_t

timfelle:
Major update to the design type.
This PR will improve the reuseability of the design type by hiding implementation details behind a generic API.


sources/designs/design.f90
```diff
@@ -1,83 +1,155 @@
...
+  !! additional setup.
+  !!
+  !! @param object The design object.
+  !! @param parameters The JSON file.
+  !! @param simulation The simulation object [Optional].
+  interface design_factory
+     module subroutine design_factory(object, parameters, simulation)
+       class(design_t), allocatable, intent(inout) :: object
+       type(json_file), intent(inout) :: parameters
+       type(simulation_t), intent(inout), optional :: simulation
```
gloopydoop, line 125: nice :+1:

sources/designs/design_brinkman.f90
```diff
@@ -389,8 +403,42 @@ subroutine topopt_design_map_forward(this)
 
   end subroutine topopt_design_map_forward
 
+  function topopt_design_get_design(this) result(x)
```
gloopydoop: for clarity, if we're switching all the `topopt_design` into `brinkman_design` we should change the names of these functions too?

sources/designs/design_simple.f90
```diff
@@ -0,0 +1,179 @@
...
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+
+! Implements the `simple_design_t` type.
+module simple_design
+  use num_types, only: rp, sp
+  use field, only: field_t
+  use json_module, only: json_file
+  use mapping, only: mapping_t
+  use PDE_filter, only: PDE_filter_t
```
gloopydoop, line 39: for the simple design I bet a lot of these could be cleaned up no?

sources/problems/objectives/lube_term_objective.f90
```diff
@@ -170,18 +172,19 @@ subroutine lube_term_init_attributes(this, design, simulation, weight, &
...
-       call neko_error('Minimum dissipation only works with topopt_design')
+       call neko_error('Minimum dissipation only works with brinkman_design')
     end select
 
     this%u => neko_field_registry%get_field('u')
     this%v => neko_field_registry%get_field('v')
     this%w => neko_field_registry%get_field('w')
-    this%B => simulation%neko_case%fluid%c_Xh%B
-    this%B_d = simulation%neko_case%fluid%c_Xh%B_d
+    this%c_Xh => simulation%neko_case%fluid%c_Xh
```
gloopydoop, line 187: ok great! I did the same thing in my template too.
So I guess rule of thumb is: Bring over the minimum amount of stuff needed, with the exception of `coef` because that one is really useful :sweat_smile:

sources/problems/objectives/minimum_dissipation_objective.f90
```diff
@@ -155,19 +168,28 @@ subroutine minimum_dissipation_init_attributes(this, design, simulation, &
...
     ! Save the simulation and design
-    this%fluid => simulation%fluid_scheme
-    this%adjoint => simulation%adjoint_case%scheme
+    this%u => neko_field_registry%get_field('u')
+    this%v => neko_field_registry%get_field('v')
+    this%w => neko_field_registry%get_field('w')
+    this%c_Xh => simulation%fluid_scheme%c_Xh
+    this%adjoint_u => neko_field_registry%get_field('u_adj')
+    this%adjoint_v => neko_field_registry%get_field('v_adj')
+    this%adjoint_w => neko_field_registry%get_field('w_adj')
```
gloopydoop, line 177: at some point we should standardize our naming conventions when it comes to the word adjoint :sweat_smile: it's a bit all over the place right now and it gets confusing.

I feel like at the beginning (ie, writing the adjoint convective term) we had a lot of suffixes with `_adj`

Now I've started to favour the prefix `adjoint_` too.

Should we lock this down as the standard convention?

gloopydoop (approved):
Looks good, pulled in and everything runs fine.
I only left a few comments

sources/designs/design_simple.f90
timfelle: Yes, very much so.

sources/problems/objectives/lube_term_objective.f90
timfelle: Well, bring over the smallest number of things over while being specific. In this case, bringing two specific arrays related to the mass matrix is not as specific as saying you need access to the integration space.

sources/problems/objectives/minimum_dissipation_objective.f90
timfelle: That would probably be a good idea. Lets discuss it tomorrow at our meeting.

sources/designs/design_brinkman.f90
timfelle: Yes, I'll do that.

PR #104, opened 2025-02-24
Avoid built-in names and handle compilation errors

timfelle:
Ensure variable names do not conflict with built-in names and exit with an error message if examples fail to compile.


gloopydoop (approved):
pulled and compiled :+1: 
tip top!

PR #122, opened 2025-03-11
Ensure cross compilation support and clean up build flags

timfelle:
Introduce a new file for compiler flags to streamline the build process and ensure compatibility with cross compilation. This change simplifies the management of compiler options for different build types.


gloopydoop (approved):
This looks good. 
Perhaps a little comment, but I guess it's not so related to this PR: We should maybe put some thought into the folder structure of `neko-top` in general. A lot of the good stuff regarding compilation/execution live in `scripts`, but then `scripts` also contains things like jupyter notebooks... Then this cmake folder only holds 2 files and sits on the top level. So it's not super intuitive what the purpose of all these top level folders are for. Maybe we should shuffle things around a bit to have a top level folder dedicated to compilation/execution and then have `scripts` less ambiguous and more related to post processing scripts. i.e. I would never expect an end user to need/want to modify/contribute a compilation script, that's more of a developer's job. Whereas I probably would want an end user to want to contribute a post processing script, so keeping these separate also makes sense to me.

Anyway, as I said, unrelated to PR. This PR is good :+1:

PR #124, opened 2025-03-12
Mapping cascade

gloopydoop:
This PR has a couple of nice features, but is still a bit of a work in progress (maybe)

- The main contribution is the notion of a mapping cascade and a lot of JSON support for the pre-existing mappings. Now we prescribe a list of mappings which are performed sequentially, and it also takes care of the chain rule moving backwards.
- There are some cosmetic changes to variable names, particularly `sens_in` and `sens_out` :sweat_smile: 
- Obviously much of the mapping was inspired by the source term handler, so in principal this could be extended to allow user defined mappings in some way, but this is to be discussed.

This PR is still in draft mode because I think we should really discuss who owns the mapping, and how we should handle mapping to multiple coeffients, and how we should hand constraints defined on intermediate steps (ie, the potential for the volume constraint to defined on the filtered design variable)

There are still a few cosmetics changes that need to be done so I'm leaving this on draft mode, but take a look if you can and tell me what you think


sources/mapping_functions/linear_mapping.f90
```diff
@@ -71,22 +72,24 @@ subroutine linear_mapping_init_from_json(this, json, coef)
...
   !> Actual constructor.
-  subroutine linear_mapping_init_from_attributes(this, coef)
+  subroutine linear_mapping_init_from_attributes(this, coef, f_min, f_max)
     class(linear_mapping_t), intent(inout) :: this
     type(coef_t), intent(inout) :: coef
+    real(kind=rp), intent(in) :: f_min, f_max
 
-    ! there's actually nothing to do here.
+    this%f_min = 0.0_rp
+    this%f_max = 1000.0_rp
```
gloopydoop: mistake, fix tomorrow morning

sources/mapping_functions/mapping_fctry.f90
```diff
@@ -0,0 +1,82 @@
...
+       allocate(PDE_filter_t::object)
+    else if (trim(type_name) .eq. "RAMP") then
+       allocate(RAMP_mapping_t::object)
+    else
+       type_string = concat_string_array(MAPPING_KNOWN_TYPES, &
+            NEW_LINE('A') // "-  ", .true.)
+       call neko_error("Unknown mapping type: " &
+            // trim(type_name) // ".  Known types are: " &
+            // type_string)
+    end if
```
timfelle: Please use the neko_type_error function instead.
```suggestion
    else
       call neko_type_error("Mapping function", type_name, MAPPING_KNOWN_TYPES)
    end if
```

sources/mapping_functions/mapping_fctry.f90
```diff
@@ -0,0 +1,82 @@
...
+    character(len=:), allocatable :: type_name
+    character(len=:), allocatable :: type_string
+
+    call json_get(json, "type", type_name)
+
+    if (trim(type_name) .eq. "linear") then
+       allocate(linear_mapping_t::object)
+    else if (trim(type_name) .eq. "PDE_filter") then
+       allocate(PDE_filter_t::object)
+    else if (trim(type_name) .eq. "RAMP") then
```
timfelle: Use a select block instead, it is much cleaner and easier to read

```
select case (trim(type_name))
case ("linear")
     ......
```

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+    ! our final mapping should now live in tmp_fld_out
+    call field_copy(X_out, tmp_fld_out)
+
+    ! free the scratch.
+    call neko_scratch_registry%relinquish_field(temp_indices)
+
+  end subroutine mapping_handler_apply_forward
+
+  !> apply the cascade of mapping_cascade.
+  !! @param X_out The mapped field ($\tilde{\rho}$)
```
timfelle: Wrong name in documentation

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+  !! @param X_in The unmapped field ($\rho$)
+  subroutine mapping_handler_apply_backward(this, sens_out, sens_in)
+    class(mapping_handler_t), intent(inout) :: this
+    type(field_t), intent(in) :: sens_in
+    type(field_t), intent(inout) :: sens_out
```
timfelle: Please match the order of inputs for the function

sources/mapping_functions/brinkman_mapping.f90
```diff
@@ -0,0 +1,93 @@
...
+
+contains
+
+  !> Constructor.
+  subroutine brinkman_mapping_init(this, coef)
+    class(brinkman_mapping_t), intent(inout) :: this
+    type(coef_t), target, intent(in) :: coef
+! This may be stupid.. It was way too heavily inspired by the source terms.
+! So Tim, double check if this was the right way of doing things or this has
+! just made the mappings overly convoluded...
```
timfelle: I don't think it is. You should not make the mapping_handler_t abstract. It is
suppose to wrap the abstract mapping types.

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+  implicit none
+  private
+
+  !> Abstract class for handling mapping_cascade.
+  !!
+  !! @details
+  !! This class is responsible for managing the mapping_cascade in a sequential
+  !! manor. It is also responsible for using the chain rule to propogate
+  !! sensitivity backwards throughout the system.
+  type, abstract, public :: mapping_handler_t
```
timfelle: Should be a complete type

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+     generic :: add => add_mapping, add_json_mappings
+     !> Append a new mapping to the mapping_cascade array.
+     procedure, pass(this) :: add_mapping => &
+          mapping_handler_add_mapping
+     !> Read from the json file and initialize the mapping_cascade.
+     procedure, pass(this) :: add_json_mappings => &
+          mapping_handler_add_json_mappings
+     !!> Initialize the user mapping.
+     ! procedure(mapping_handler_init_user_mapping), &
+     !     nopass, deferred :: init_user_mapping
```
timfelle: We are not going with the user based injection. Thats what the "add_mapping"
function is for

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+          end if
+       else
+          i0 = 0
+          allocate(this%mapping_cascade(n_mappings))
+       end if
+
+       do i = 1, n_mappings
+          ! Create a new json containing just the subdict for this mapping.
+          call json_extract_item(json, name, i, mapping_subdict)
+          call json_get(mapping_subdict, "type", type)
```
timfelle: Why do you grab the type. You never use it.

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,291 @@
...
+       n_mappings = 0
+    end if
+
+    call move_alloc(this%mapping_cascade, temp)
+    allocate(this%mapping_cascade(n_mappings + 1))
+
+    if (allocated(temp)) then
+       do i = 1, n_mappings
+          call move_alloc(temp(i)%mapping, &
+               this%mapping_cascade(i)%mapping)
```
timfelle: Why is this line broken? it is not too long

sources/mapping_functions/mapping.f90
```diff
@@ -37,31 +37,44 @@ module mapping
...
+     procedure, pass(this) :: apply_backward => mapping_apply_backward_wrapper
      !> The common constructor using a JSON dictionary.
      procedure(mapping_init), pass(this), deferred :: init
      !> Destructor.
      procedure(mapping_free), pass(this), deferred :: free
      !> Apply forward
-     procedure(mapping_apply), pass(this), deferred :: apply_forward
-     !> Apply backwards (with chain rule)
-     procedure(mapping_apply_backward), pass(this), deferred :: apply_backward
+     procedure(mapping_apply), pass(this), deferred :: apply_forward_
```
timfelle: If you do not intend the user to use the "function_name_" then please make it
private. Convoluted names like that should only be used for internal functions
that you know no-one other than yourself will ever use.

sources/mapping_functions/mapping_cascade.f90
timfelle, line 1: Please remove unused files

sources/mapping_functions/brinkman_mapping.f90
timfelle, line 1: What is the purpose of this file? It does not seem useful

sources/mapping_functions/CMakeLists.txt
timfelle, line 1: Please fix the cmake version command here. That is not suppose to be here. Also
don't set the cuda_architecture. As we learned on the lumi course, it prevents
cross compilation.

sources/neko_ext/neko_ext.f90
```diff
@@ -180,4 +187,52 @@ subroutine setup_iteration(neko_case, iter)
...
+  !! that they have the same dimension.
+  !!
+  !! @param[out] vector the output vector.
+  !! @param[in] field the input field.
+  subroutine field_to_vector(vector, field)
+    type(vector_t), intent(inout) :: vector
+    type(field_t), intent(in) :: field
+
+    ! first check they're the same size
+    if (field%size() .ne. vector%n) then
```
timfelle: A vector also have a size, please use that and not the internal components.

sources/neko_ext/neko_ext.f90
```diff
@@ -180,4 +187,52 @@ subroutine setup_iteration(neko_case, iter)
...
+  !! that they have the same dimension.
+  !!
+  !! @param[out] field the output field.
+  !! @param[in] vector the input vector.
+  subroutine vector_to_field(field, vector)
+    type(field_t), intent(inout) :: field
+    type(vector_t), intent(in) :: vector
+
+    ! first check they're the same size
+    if (field%size() .ne. vector%n) then
```
timfelle: A vector also have a size, please use that and not the internal components.

timfelle (changes requested):

sources/mapping_functions/CMakeLists.txt
gloopydoop, line 1: remove this

gloopydoop: plz double check that this is correct now. (note, it was copied from `neko_ext/math/CMakeList.txt` so I've made the changes there too.

This week I don't currently have access to a GPU for testing, so I really need you to double check this :+1:

sources/mapping_functions/brinkman_mapping.f90
gloopydoop: so much cleaner :+1:

sources/neko_ext/neko_ext.f90
gloopydoop: this one is fine with `vector%size()`

sources/neko_ext/neko_ext.f90
gloopydoop: this one gives me:
`Error: Dummy argument ‘vector’ with INTENT(IN) in variable definition context (actual argument to INTENT = OUT/INOUT)`

I really want to keep the `intent(in)` for these subroutines

timfelle: That is a bug in Neko. Please submit a PR with a fix. Otherwise i can do it.

sources/mapping_functions/mapping.f90
timfelle: This is still not addressed

sources/neko_ext/neko_ext.f90
timfelle: Then use it

timfelle (changes requested):
There are still old conversations that have not been addressed

sources/neko_ext/neko_ext.f90
gloopydoop: on it :+1:

sources/mapping_functions/PDE_filter_mapping.f90
```diff
@@ -124,26 +125,38 @@ subroutine PDE_filter_init_from_json(this, json, coef)
...
-    this%precon_type_filt = "jacobi"
+    call json_get(json, 'r', r)
+    call json_get_or_default(json, 'tol', tol, 0.0000000001_rp)
+    call json_get_or_default(json, 'max_iter', max_iter, 200)
+    call json_get_or_default(json, 'solver', ksp_solver, "cg")
+    call json_get_or_default(json, 'preconditioner', precon_type, "jacobi")
 
     call this%init_base(json, coef)
-    call PDE_filter_init_from_attributes(this, coef)
+    call PDE_filter_init_from_attributes(this, coef, r, tol, max_iter, &
```
timfelle:
```suggestion
    call this%init_from_attributes(coef, r, tol, max_iter, &
```

sources/mapping_functions/PDE_filter_mapping.f90
```diff
@@ -328,20 +341,20 @@ subroutine PDE_filter_apply_backward(this, dF_dX_in, dF_dX_out, X_in)
...
+       call device_subcol3(RHS%x_d, sens_in%x_d, this%coef%B_d, n)
        call device_cmult(RHS%x_d, -1.0_rp, n)
     else
        do i = 1, n
           ! mass matrix should be included here
-          RHS%x(i,1,1,1) = dF_dX_out%x(i,1,1,1) * this%coef%B(i,1,1,1) &
+          RHS%x(i,1,1,1) = sens_in%x(i,1,1,1) * this%coef%B(i,1,1,1) &
                - RHS%x(i,1,1,1)
        end do
     end if
```
timfelle, line 360: Why not use the equivalent operators for the CPU variant?

sources/mapping_functions/RAMP_mapping.f90
```diff
@@ -107,24 +108,32 @@ subroutine RAMP_mapping_init_from_json(this, json, coef)
...
-    this%q = 1.0_rp
-    this%convex_up = .true.
+    call json_get_or_default(json, 'f_min', f_min, 0.0_rp)
+    call json_get(json, 'f_max', f_max)
+    call json_get_or_default(json, 'q', q, 1.0_rp)
+    call json_get_or_default(json, 'convex_up', convex_up, .true.)
 
     call this%init_base(json, coef)
-    call RAMP_mapping_init_from_attributes(this, coef)
+    call RAMP_mapping_init_from_attributes(this, coef, f_min, f_max, q, &
```
timfelle:
```suggestion
    call this%init_from_attributes(coef, f_min, f_max, q, &
```

sources/mapping_functions/PDE_filter_mapping.f90
gloopydoop: one less loop.

But I was actually thinking about this when the compiler guy was talking at the LUMI course about sometime the compiler is smart enough to merge loops together. Maybe it would for the CPU stuff, but I doubt it for the GPU stuff. It would be cool functionality in `neko` to be able to merge kernels together somehow.
to have something like
```fortran
merge
call device_subcol3(RHS%x_d, sens_in%x_d, this%coef%B_d, n)
 call device_cmult(RHS%x_d, -1.0_rp, n)
end merge
```
or something to that effect, and it would only loop once and merge the internal kernels

sources/mapping_functions/mapping.f90
gloopydoop: still working on this one, specifically the `apply_backward`.
I want the internal `apply_backward` (what I was calling `apply_backward_`) to take three arguments, sens_out, sense_in, X_in

but I want `apply_backward` to only take two, sens_out, sense_in, and have it place the internally saved X_in into the third spot

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,276 @@
...
+  type, public :: mapping_handler_t
+     !> Array of mapping_cascade.
+     !! @note the order really matter's here since they'll be executed in
+     !! sequence.
+     class(mapping_wrapper_t), allocatable :: mapping_cascade(:)
+     !> The right-hand side.
+     type(field_list_t) :: rhs_fields
+     !> The coefficients of the (space, mesh) pair.
+     type(coef_t), pointer :: coef
+     ! @todo, we should handle user mapping_cascade in a different PR
```
timfelle: I guess we did

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,276 @@
...
+     !!> The user object.
+     !!type(user_t), pointer :: user
```
timfelle: Please remove this

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,276 @@
...
+       end do
+    else
+       ! I was considering an error, but maybe a warning is more appropriate
+       call neko_warning("No mappings selected")
+    end if
+
+  end subroutine mapping_handler_add_json_mappings
+
+  !> Add new mapping to the list.
+  !! @param mapping The mapping to be added.
```
timfelle, line 247: You are missing `!! @params this`, If you document the parameters, you should
document all of them.

sources/mapping_functions/mapping_handler.f90
```diff
@@ -0,0 +1,276 @@
...
+       do i = 1, size(this%mapping_cascade)
+          call this%mapping_cascade(i)%free()
+       end do
+       deallocate(this%mapping_cascade)
+    end if
+
+  end subroutine mapping_handler_free
+
+  !> apply the cascade of mapping_cascade.
+  !! @param X_out The mapped field ($\tilde{\rho}$)
```
timfelle: Please fix your equations. the doxygen construction for latex equations are \f$
EQ \f$

sources/mapping_functions/mapping.f90
timfelle, line 1: Please fix the math structures in the doxygen comments. Inline math is done by: `\f$ EQ \f$`

timfelle (changes requested):
Please go through and fix the Doxygen math commands.
Inline math: `\f$ EQ \f$`
Math blocks `\f[ EQ \f]`

timfelle (approved):
Scrap that, we can do the doxygen in a separate PR. But keep it in mind going forward.

sources/mapping_functions/mapping.f90
```diff
@@ -121,16 +148,61 @@ subroutine mapping_init_base(this, json, coef)
...
+    end if
+  end subroutine mapping_wrapper_free
+
+  !> Executes `apply_forward_`
+  !! @param X_out The mapped field ($\tilde{\rho}$)
+  !! @param X_in The unmapped field ($\rho$)
+  subroutine mapping_apply_forward_wrapper(this, X_out, X_in)
+    class(mapping_t), intent(inout) :: this
+    type(field_t), intent(in) :: X_in
+    type(field_t), intent(inout) :: X_out
```
timfelle: Please be consistent in the ordering of your inputs and outputs

sources/mapping_functions/mapping_handler.f90
gloopydoop: I don't think `this` is documented as a parameter anywhere else in neko

sources/mapping_functions/mapping_handler.f90
gloopydoop: resolved, but very good comment :+1:

PR #125, opened 2025-03-13
Sync formatting to neko standards

timfelle:
Update formatting commands to align with neko's specifications.

Continuation lines starting with a `&` is not indented to 0 compared to their current scope.

This is commonly used to break long strings.


gloopydoop (approved):
I was semi lurking on your conversation on zulip :sweat_smile:  :+1:

gloopydoop:
Is this going to mean we need to findent the whole source? (I actually assumed this PR was going to fail on the linting and you only had it open to see where it fails :sweat_smile: )

PR #127, opened 2025-03-14
Initial material distribution

gloopydoop:
This one is pretty self explanatory.

But let me know your thoughts on moving the boundary_mesh utilities out of the `brinkman_source_term` and into `set_scalar_ic` in `neko`, and then we all have public access to that functionality.

I guess something also up for discussion is how multiple point zones is handled. Currently it's quite nice in the `brinkman_source_term` but I also saw that `point_zones` support boolean operations, so maybe this is the way forward.

https://neko.cfd/docs/develop/da/dd0/point-zones.html

Depends on #124


timfelle:
@gloopydoop Did you mean to target the mapping PR?

gloopydoop:
> @gloopydoop Did you mean to target the mapping PR?

I did because I'd already pulled that branch into this branch and the files changed looked cleaner this way :sweat_smile: . Both of them are still darft, I have some final testing and tinkering to do. I most made this PR to see if you agree with moving a lot of the `brinkman_source_term` stuff over to `scalar_ic` on the `neko` side such that those routines are a bit more public. The alternative is we write a brand new set of routines for setting the initial material distribution but that feels like a lot of double of code.

On a side note, at some point we need to write a `neko-top` equivalent of `user_inf.f90`, I think user defined initial material distributions would be a valuable addition.

sources/designs/design_brinkman.f90
```diff
@@ -214,16 +229,33 @@ subroutine brinkman_design_init_from_json_sim(this, parameters, simulation)
...
+      call json_extract_object(parameters, &
+           'optimization.design.initial_distribution', json_subdict)
+
+      if (trim(string_val) .ne. 'user') then
+         call set_optimization_ic(this%design_indicator, coef, gs, string_val, &
+            json_subdict)
+      else
+         call neko_error("user defined initial material distrubtions should &
+         & be enforced through the driver.")
+      end if
```
timfelle: We will not spent time implementing a user system similar to the one in neko.
There is no benefit from doing so.
Also why are you passing the string_value in? The exact information is also
contained in the json_subdict.
```suggestion
      call set_optimization_ic(this%design_indicator, coef, gs, string_val, &
           json_subdict)
```

gloopydoop: take another look, I figured we should also start with zero if nothing is prescribed

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,373 @@
...
+    character(len=:), allocatable :: type
+    real(kind=rp) :: ic_value
+    character(len=:), allocatable :: read_str
+    character(len=NEKO_FNAME_LEN) :: fname, mesh_fname
+    real(kind=rp) :: zone_value, tol
+    logical :: interpolate
+
+    call json_get(params, "type", type)
+
+    if (trim(type) .eq. 'uniform') then
```
timfelle: Please use `select case` for this kind of switching, it is a lot cleaner.

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,373 @@
...
+       call json_get(params, 'base_value', ic_value)
+       call json_get(params, 'zone_name', read_str)
+       call json_get(params, 'zone_value', zone_value)
+
+       call set_optimization_ic_point_zone(fld, ic_value, read_str, zone_value)
+
+    else if (trim(type) .eq. 'field') then
+
+       call json_get(params, 'file_name', read_str)
+       fname = trim(read_str)
```
timfelle: How about trimming inside the function instead, then you also safeguard other
users. and you can read directly into thevariable

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,373 @@
...
+       call json_get_or_default(params, 'interpolate', interpolate, &
+            .false.)
+       call json_get_or_default(params, 'tolerance', tol, 0.000001_rp)
+       call json_get_or_default(params, 'mesh_file_name', read_str, &
+            "none")
```
timfelle: Does this not fit in one line?

sources/neko_ext/neko_ext.f90
```diff
@@ -24,6 +24,14 @@ module neko_ext
...
   use json_module, only : json_file
+  ! for vector/field math
+  use math, only: copy
+  use device_math, only: device_copy
+  use neko_config, only : NEKO_BCKND_DEVICE
+  use vector, only: vector_t
+  use field, only: field_t
+  use utils, only: neko_error
+  use json_module, only : json_file
   implicit none
```
gloopydoop: I don't know how #124 got through without this? I didn't do the best job of splitting things apart :sweat_smile:

sources/designs/set_optimization_ic.f90
gloopydoop: I'm not sure I understand completely, and it was Victor who implemented `set_scalar_IC` so maybe I can talk to him when I get back to Stockholm, but those two names were declared rather deliberatly as:
```fortran 
character(len=NEKO_FNAME_LEN) :: fname, mesh_fname
``` 
so he must have had his reasons

timfelle: My suggestion is, remove trim here

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,371 @@
...
+    ! Extract sample index from the file name
+    sample_idx = extract_fld_file_index(file_name, -1)
+
+    if (sample_idx .eq. -1) then
+       call neko_error("Invalid file name for the initial condition. The " // &
+            "file format must be e.g. 'mean0.f00001'")
+    end if
+
+    ! Change from "field0.f000*" to "field0.fld" for the fld reader
+    call filename_chsuffix(file_name, file_name, 'fld')
```
timfelle: Trim here instead

examples/rugby_ball/rugby_ball.case
```diff
@@ -86,7 +86,14 @@
...
                 ],
-            }
+            },
+            {
+                "name": "initial_ball",
+                "geometry": "cylinder",
+                "start": [0.5, 0.5, -0.5],
+                "end": [0.5, 0.5, 1.5],
+                "radius": 0.25
+    },
```
timfelle:
```suggestion
            {
                "name": "initial_ball",
                "geometry": "cylinder",
                "start": [0.5, 0.5, -0.5],
                "end": [0.5, 0.5, 1.5],
                "radius": 0.25
            },
```

examples/rugby_verification/rugby_ball.case
```diff
@@ -81,7 +81,14 @@
...
                 ],
-            }
+            },
+            {
+                "name": "initial_ball",
+                "geometry": "cylinder",
+                "start": [0.5, 0.5, -0.5],
+                "end": [0.5, 0.5, 1.5],
+                "radius": 0.25
+    },
```
timfelle:
```suggestion
            {
                "name": "initial_ball",
                "geometry": "cylinder",
                "start": [0.5, 0.5, -0.5],
                "end": [0.5, 0.5, 1.5],
                "radius": 0.25
            },
```

sources/designs/design_brinkman.f90
```diff
@@ -227,16 +231,29 @@ subroutine brinkman_design_init_from_json_sim(this, parameters, simulation)
...
 
     ! Initialize the mapper
-    associate(coef => simulation%neko_case%fluid%c_Xh)
+    associate(coef => simulation%neko_case%fluid%c_Xh, &
+         gs => simulation%neko_case%fluid%gs_Xh)
       call this%mapping%init_base(coef)
       call this%mapping%add(parameters, 'optimization.design.mapping')
+
+      if (parameters%valid_path(&
+           'optimization.design.initial_distribution')) then
```
timfelle, line 267: Lets make this mandatory and allow the user to specify a "uniform" distribution
as well

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,371 @@
...
+    real(kind=rp), intent(in) :: base_value
+    character(len=*), intent(in) :: zone_name
+    real(kind=rp), intent(in) :: zone_value
+
+    ! Internal variables
+    character(len=LOG_SIZE) :: log_buf
+    class(point_zone_t), pointer :: zone
+    integer :: size
+
+    call neko_log%message("Type       : point_zone")
```
timfelle: Why all the spaces after type? it does not even match with the number of
characters in the "Base value" line just under it.

sources/designs/set_optimization_ic.f90
```diff
@@ -0,0 +1,371 @@
...
+    else ! No interpolation, just potentially from different spaces
+
+       ! Build a space_t object from the data in the fld file
+       call prev_Xh%init(GLL, fld_data%lx, fld_data%ly, fld_data%lz)
+       call space_interp%init(fld%Xh, prev_Xh)
+
+       ! Do the space-to-space interpolation
+       call space_interp%map_host(fld%x, fld_data%t%x, fld_data%nelv, fld%Xh)
+
+       call space_interp%free
```
timfelle: Please don't blindly copy Neko. 
Add parenthesis for function calls, even when no parameters are given.

timfelle (changes requested):

timfelle (approved):

PR #129, opened 2025-03-17
Align with new KSP in neko

timfelle:
Update the usage of the Krylov solver to align with the new KSP interface, replacing calls to `krylov_solver_destroy` with the appropriate `free` method for memory management.


gloopydoop (approved):
:+1:

PR #110, opened 2025-02-28
Passive scalar support

gloopydoop:
This PR contains:

-  Adjoint passive scalar support
- A scalar mixing objective function
- An example of a mixer

PR #134, opened 2025-03-25
Fix segmentation fault and improve logging

timfelle:
Address segmentation faults by ensuring proper memory management and initialization. Enhance logging capabilities for better debugging and add tests to verify functionality. Align code with new checkpoint structures and clean up unnecessary outputs.

### DIFF
--- a/.github/workflows/micro_test.yml
+++ b/.github/workflows/micro_test.yml
@@ -63,5 +63,6 @@ jobs:
         run: |
           ./run.sh -rc easy-E/easy-E_1 rugby_ball
           if [ -d logs ]; then
+            find logs/ -type f -name "*.log" -exec echo {}; cat {} \;
             exit 1
           fi
--- a/.github/workflows/micro_test.yml
+++ b/.github/workflows/micro_test.yml
@@ -75,5 +75,7 @@ jobs:
               echo "End of log file: $log"
             done
 
+            ls -l logs/rugby_ball
+
             exit 1
           fi
--- a/.github/workflows/micro_test.yml
+++ b/.github/workflows/micro_test.yml
@@ -66,8 +66,13 @@ jobs:
           if [ -d logs ]; then
             LOGS=$(find logs/ -type f -name "*.log")
             for log in ${LOGS[@]}; do
+              echo ""
+              echo ""
               echo "Log file: $log"
+              echo ""
               cat $log
+              echo ""
+              echo "End of log file: $log"
             done
 
             exit 1
--- a/.github/workflows/micro_test.yml
+++ b/.github/workflows/micro_test.yml
@@ -6,6 +6,10 @@ name: Test Scripts
 # This workflow is only defined here. It is triggered from other workflows.
 on:
   push:
+    branches:
+      - test/*
+      - cicd/*
+  workflow_dispatch:
   merge_group:
     types: [checks_requested]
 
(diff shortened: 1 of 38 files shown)


sources/adjoint/adjoint_case.f90
```diff
@@ -168,8 +169,8 @@ subroutine adjoint_case_init_common(this, neko_case)
     json_key = json_key_fallback(neko_case%params, &
          'case.adjoint_fluid.initial_condition', 'case.fluid.initial_condition')
 
-    call json_get(neko_case%params, json_key//'.type', string_val)
     call json_extract_object(neko_case%params, json_key, ic_json)
+    call json_get(ic_json, 'type', string_val)
```
gloopydoop, line 173: this was it?!?
Or at least, can you shed a little light on the change, because I'll need to update #110 too

sources/simulation/simulation.f90
```diff
@@ -85,12 +86,24 @@ module simulation
...
+  end subroutine user_simcomp
+
   !> Initialize the simulation
   subroutine simulation_init(this, parameters)
     class(simulation_t), intent(inout) :: this
     type(json_file), intent(inout) :: parameters
-    type(steady_simcomp_t), allocatable :: steady_comp
-    type(json_file) :: simcomp_settings
+
+    this%neko_case%usr%init_user_simcomp => user_simcomp
```
gloopydoop, line 106: so in #110 for the forward problem I used a lot of built `neko` user interface for the forward problem.
I guess here we're forcing a simulation to be steady, perhaps what would be more natural would be to include this in the drivers of the examples to indicate we're doing a steady simulation

gloopydoop (commented):
This looks good! 
I've put some questions because I encountered similar problems in neko #1738 and couldn't put my finger on why `json_extract_object` was necessary. 

I'll click approve tomorrow after I've tested a little, but this looks good :+1:

sources/adjoint/adjoint_case.f90
timfelle: No, the problem was in neko itself and how it handled the user simulation components. Which i have opened a pr fixing on the neko side.

sources/simulation/simulation.f90
timfelle: Let's discuss this on the actual pr instead.

But no, I do not agree. The user simcomp here is a temporary hack to make things work, and have always been that. A steady "solve" function should be implemented in neko. And this will in any case become a "steady state simulation" to distinguish it from an "unsteady simulation", which we have not yet implemented.

And another quick note. User interfaces are for example by example code and will be overwritten by users. So please do not use it in the src folder. It is only here as a hack for the steady and i will remove it as soon as the code compiles and runs with the bugs fixed.

.github/workflows/micro_test.yml
```diff
@@ -0,0 +1,76 @@
...
+              echo ""
+              echo "Log file: $log"
+              echo ""
+              cat $log
+              echo ""
+              echo "End of log file: $log"
+            done
+
+            exit 1
+          fi
```
gloopydoop, line 76: is this maybe a part of #107 ?

timfelle: No, this is just adding a quick sanity check to prevent the segfault issue from arising again.

.github/workflows/micro_test.yml
```diff
@@ -0,0 +1,76 @@
+# Description: This file defines the workflow that will be run when the a new
```
gloopydoop:
```suggestion
# Description: This file defines the workflow that will be run when a new
```

gloopydoop (approved):

PR #138, opened 2025-03-27
Improve documentation and fix formatting issues

timfelle:
Enhance the documentation configuration and deployment, correct markdown formatting, and complete doxygen comments across multiple files. Addressed bugs in doxygen and made spelling updates for clarity.

### DIFF
--- a/.github/workflows/deploy_documentation.yml
+++ b/.github/workflows/deploy_documentation.yml
@@ -31,9 +31,12 @@ jobs:
           packages: git cmake doxygen graphviz
           version: 1.0
 
+      - name: Configure documentation
+        run: |
+            cmake -DONLY_DOCS=ON -B build -S .
+
       - name: Build documentation
         run: |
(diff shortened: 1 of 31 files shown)


gloopydoop (commented):
wow this is a lot :+1: 

Q1: So exactly what is being checked in the `deploy_documentation.yml` on github if this PR goes through. Just equations? or also `@param` stuff?

Q2: What's the protocol with `@param this`, I was searching through all of `neko` and no one does it.

timfelle (commented):
Deploy documentation checks nothing, it only deploys the documentation once a push is done to `develop`.

`check_documentation.yml` does a test build of the documentation. In that, all warnings will be treated as errors. Which mean, unsupported commands in doxygen, such as `\rho` not in a math environment will throw an error. Additionally, if any function parameters are documented with `@params`, then all the parameters of the function should be documented.

gloopydoop (approved):
allright well let's give it a crack and see what happens

PR #139, opened 2025-03-27
Optimization domain in JSON

gloopydoop:
Wow I can't believe this one was still hardcoded :sweat_smile: 
I've put the "help wanted" flag on because I don't know how this translates into an optimization problem without a simulation. I guess the word "domain" for topology optimization means a spatial domain, so to me I think it should probably be rooted in `optimization.design.domain`

But it's worth discussing.

Also aiming for many small PRs moving forward :sweat_smile:

### DIFF
--- a/examples/rugby_ball/rugby_ball.case
+++ b/examples/rugby_ball/rugby_ball.case
@@ -93,7 +93,7 @@
         "domain": {
             "type": "point_zone",
             "zone_name": "optimization_domain"
-            },
+        },
         "design": {
             "type": "brinkman",
             "mapping" : [
--- a/examples/rugby_ball/rugby_ball.case
+++ b/examples/rugby_ball/rugby_ball.case
@@ -91,9 +91,9 @@
     },
     "optimization": {
         "domain": {
-                    "type": "point_zone",
-                    "zone_name": "optimization_domain"
-                  },
+            "type": "point_zone",
+            "zone_name": "optimization_domain"
+            },
         "design": {
             "type": "brinkman",
             "mapping" : [
--- a/examples/rugby_ball/rugby_ball.case
+++ b/examples/rugby_ball/rugby_ball.case
@@ -90,6 +90,10 @@
         ]
     },
     "optimization": {
+        "domain": {
+                    "type": "point_zone",
+                    "zone_name": "optimization_domain"
+                  },
         "design": {
             "type": "brinkman",
             "mapping" : [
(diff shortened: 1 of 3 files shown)


examples/rugby_ball/rugby_ball.case
```diff
@@ -90,6 +90,10 @@
         ]
     },
     "optimization": {
+        "domain": {
+                    "type": "point_zone",
+                    "zone_name": "optimization_domain"
+                  },
```
timfelle: Please keep formatting the same as the rest of the file.

examples/rugby_verification/rugby_ball.case
```diff
@@ -85,6 +85,10 @@
         ]
     },
     "optimization": {
+        "domain": {
+                    "type": "point_zone",
+                    "zone_name": "optimization_domain"
+                  },
```
timfelle: Again here with the formatting

sources/designs/design_brinkman.f90
```diff
@@ -227,6 +229,28 @@ subroutine brinkman_design_init_from_json_sim(this, parameters, simulation)
     class(brinkman_design_t), intent(inout) :: this
     type(json_file), intent(inout) :: parameters
     type(simulation_t), intent(inout) :: simulation
+    character(len=:), allocatable :: optimization_domain_name, domain_type
```
timfelle: why not call it `domain_name` you will probably not have anything else called
that in this function

timfelle (commented):
The brinkman based design does not really make sense without a simulation so that part does not matter so much.

examples/rugby_ball/rugby_ball.case
```diff
@@ -90,6 +90,10 @@
         ]
     },
     "optimization": {
+        "domain": {
+            "type": "point_zone",
+            "zone_name": "optimization_domain"
+            },
```
timfelle:
```suggestion
        },
```

examples/rugby_verification/rugby_ball.case
```diff
@@ -85,6 +85,10 @@
         ]
     },
     "optimization": {
+        "domain": {
+            "type": "point_zone",
+            "zone_name": "optimization_domain"
+            },
```
timfelle:
```suggestion
        },
```

examples/rugby_ball/rugby_ball.case
gloopydoop: how did VScode let me down?!?!

timfelle (approved):

PR #140, opened 2025-03-27
Add option to build documentation

timfelle:
Introduce a command-line option to build documentation during the compilation process. This enhancement allows users to generate documentation as part of the build workflow.

Depends on: #120


gloopydoop (approved):
for clarity, is it:
`./setup.sh --doc` and then we get the same output as we see on the github website?

gloopydoop:
> for clarity, is it: `./setup.sh --doc` and then we get the same output as we see on the github website?

I think that's what I'm seeing :sweat_smile:

PR #141, opened 2025-03-27
Make simulation optional in Problem and Optimizer

timfelle:
This PR now contain the refactor to make a simulation optional in a given optimizer and problem setup.
Additionally we guard against some missing JSON keys to make sure an empty problem is supported for the user to programmatically construct the problem.

### DIFF
--- a/sources/optimizer/mma_optimizer.f90
+++ b/sources/optimizer/mma_optimizer.f90
@@ -66,14 +66,14 @@ contains
 
   !> Initialize the MMA optimizer from JSON file
   subroutine mma_optimizer_init_from_json(this, parameters, problem, design, &
-       simulation, max_iterations, tolerance)
+       max_iterations, tolerance, simulation)
     class(mma_optimizer_t), intent(inout) :: this
     type(json_file), intent(inout) :: parameters
     class(problem_t), intent(in) :: problem
     class(design_t), intent(in) :: design
-    type(simulation_t), intent(in) :: simulation
     integer, intent(in) :: max_iterations
     real(kind=rp), intent(in) :: tolerance
+    type(simulation_t), optional, intent(in) :: simulation
 
     character(len=1024) :: optimization_header
     character(len=1024) :: problem_header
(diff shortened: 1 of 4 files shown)


timfelle:
I am not exactly sure the new MMA exampe provide the expected output since it is not defined on the same mesh as the previous version. Which is why it is not overwritten yet.

examples/new_mma/mma.case
```diff
@@ -0,0 +1,35 @@
+{
+    "version": 1.0,
+    "optimization": {
+        "design": {
+            "type": "simple",
+            "domain": {
+                "type": "box",
+                "limits": [ 0.0, 10.0, 0.0, 10.0, 0.0, 10.0 ],
```
gloopydoop: I guess we have a meeting in an hour so we can discuss then, but for a `"simple"` design are we locked into have a sense of spatial coordinates? Would it not make more sense to just assign a vector and a size?

examples/new_mma/mma.case
```diff
@@ -0,0 +1,35 @@
...
+            "mma": {
+                "max_iter": 100,
+                "backend": "cpu",
+                "asyinit": 0.5,
+                "asyincr": 1.2,
+                "asydecr": 0.7,
+                "scale": 1.0,
+                "auto_scale": false,
+                "xmin": 0.00,
+                "xmax": 10.0,
```
gloopydoop: maybe for a different PR, but these should also be vectors

examples/new_mma/mma.case
```diff
@@ -0,0 +1,35 @@
...
+                "a0": 1.0,
+                "a": 0.0,
+                "c": 10000.0,
```
gloopydoop: and these

gloopydoop (commented):
ahhh I see what happened here. Ok I think this is the main comment for this PR:
I think when we came up with this test problem/objective we wanted to come up with something simple but non-trivial and so we thought $\sum (x-x_{location})^2$ was simple to implement.

It would be my opinion that a `simple` design type doesn't need to know their spatial coordinates. This seems like a more specific derived type that has awareness of spatial coordinates. So maybe we should think the naming `simple` used in this PR. Maybe we change it to something else, and reserve the name `simple` for something that really only contains a vector.

This being said, if we want to piggy back on the partitioning used in `neko` we would need to have a mesh in some sense. I don't really want to reinvent the wheel here, it would be nice to take all the partitioning used in `neko` and be done with it. We just need to be aware that the partitioning in `neko` is very specific to the spatial location of the elements in the domain. i.e. exploit the knowledge that most communication for direct stiffness summation happens across the element interfaces, so it tries to blob elements together and minimize the "surface area" between them (probably the wrong nomenclature :sweat_smile: )
This is not necessarily the case for an abstract MMA problem, so we may end up with less than ideal partitioning for an abstract problem.

Nevertheless, I think working on partitioning is not the best use of our time, and definitely beyond the scope of the project right now, so I do agree we need to hijack the mesh partitioning in `neko`. I just don't think a `simple` design type needs to know about spatial coordinates.

AbbasMoosaV (commented):
I think the changes to optimizer and problem types to make simulation optional look good.
The thing is the current mma testcase (examples/mma) does not use optimizer type. It is directly working with the mma_t (which it should not do :))))).
After finalizing the mma implementation (adding cuda support for the dual subsolver) I will revise the mma example such that it works properly through optimizer_t and mma_optimizer_t.
When we do that, then we can use this capability of having an optional simulation in the optimizer type for the mma example.

For now, I think this PR is good to go.
@timfelle , let me know if you think we need to discuss anything further before merging. Otherwise, I’ll approve and add it to the merge queue.

AbbasMoosaV (approved):

PR #135, opened 2025-03-25
Update to support pr 1727 of Neko

timfelle:
Changes tot the time structure in Neko require us to update our interfaces.

The neko work on this started with [PR #1727](https://github.com/ExtremeFLOW/neko/pull/1727), but have been superseeded by [PR #1803](https://github.com/ExtremeFLOW/neko/pull/1803)

PR #144, opened 2025-04-01
Split functional type initializer for simulation handling

timfelle:
Refactor the functional type initializer to separate initialization with and without simulation, enhancing clarity and usability. Update related modules to accommodate optional simulation parameters.


gloopydoop (approved):
This looks nice too! (I haven't tested it because it relied on #135)
But it seems very reasonable.

PR #145, opened 2025-04-02
Update cache key generation for libraries

timfelle:
Revise the cache key generation process to include version, OS, compiler, and compiler flags for multiple libraries, enhancing cache management and efficiency.

Additionally a sha value of the commit is added to mirror updates to dependencies.

Depends on: #135


gloopydoop (approved):
:+1:

PR #147, opened 2025-04-02
Refactor design object and add coordinate retrieval functions

timfelle:
Refactor the design object to improve structure and introduce functions for retrieving x, y, and z coordinates of design variables. This enhances the usability of the design module.

### DIFF
--- a/sources/designs/design.f90
+++ b/sources/designs/design.f90
@@ -36,6 +36,8 @@ module design
   use simulation_m, only: simulation_t
   use vector, only: vector_t
   use utils, only: neko_error
+  use comm, only: neko_comm
+  use mpi_f08, only: MPI_Allreduce, MPI_INTEGER, MPI_SUM
   implicit none
   private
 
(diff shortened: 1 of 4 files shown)


gloopydoop (commented):
I'm going to tag @AbbasMoosaV  in this one for his opinion.

My opinion is the following:
I don't think the base class in `design.f90` should have the deferred procedures `get_x` etc. because it forces our base class to be spatially aware, which it shouldn't have to be.

Regarding the derived type `design_simple.f90`, I actually also think the simplest version would also only contain the vector `values` (previously referred to as `x`) and the getter. Making these changes, in my opinion, worthy of an additional derived type of spatially aware simple designs.

However, maybe we want our `design_simple` to be spatially aware? That's worth discussing.

But my main point is that I really don't think the base class should have deferred procedures that force it to be spatially aware. So I would suggest removing that from `design.f90` and have them as extra procedures only in the derived types if necessary.

gloopydoop (approved):
spoke over zoom :+1:

PR #148, opened 2025-04-02
Introduce a simulation-less MMA example and improve termination handling

timfelle:
Create a new example for MMA that does not rely on simulations and enhance the termination process by removing the previous normal end check.

Depends on: #141, #147


timfelle:
@AbbasMoosaV The solution for the mma example in this PR is never feasible. The constraint stays positive throughout the optimization, but it should be negative in order to be feasible

timfelle:
Superseeded by PR #284

PR #169, opened 2025-04-22
Lumi/submit

timfelle:
Submission to the LUMI cluster through the run script and `sbatch` is now supported.


run.sh
```diff
@@ -303,10 +303,21 @@ function Submit() {
         fi
         sbatch -A $MN5_ACCOUNT -J $1 job_script.sh 1>/dev/null 2>error.log
 
+    elif [ $CLUSTER == "LUMI" ]; then
```
gloopydoop, line 306: let's specifically state "LUMI-C" here, because I'm working on the "LUMI-G" part and there's lots of extra steps regarding the CPU binding etc that we'll need in those submission scripts

timfelle: So you want to associate two different jobscripts to the same example? In order to have a CPU and a GPU version of the jobscript? How do you guarantee that the jobscript of the LUMI-G version is actually associated with the G partition? And do you also want a LUMI-small for the small partition?

gloopydoop: Yeh I think distinguishing between CPU and GPU job scripts is a good idea. Here's an example of a jobscript for running on LUMI that I got from Bertie:

```bash
#!/bin/bash -l
#SBATCH --job-name=passive_scalar # Job name
#SBATCH --output=log.o%j         # Name of stdout output file
#SBATCH --error=log.e%j          # Name of stderr error file
#SBATCH --partition=standard-g   # Partition (queue) name
#SBATCH --nodes=1                # Total number of nodes 
#SBATCH --ntasks-per-node=8      # 8 MPI ranks per node
#SBATCH --gpus-per-node=8        # Allocate one gpu per MPI rank
#SBATCH --time=0-00:05:00        # Run time (d-hh:mm:ss)
#SBATCH --mail-type=all          # Send email at begin and end of job
#SBATCH --account=project_465001539  # Project for billing
#SBATCH --mail-user=nobis@kth.se

cat << EOF > select_gpu
#!/bin/bash

export ROCR_VISIBLE_DEVICES=\$SLURM_LOCALID
exec \$*
EOF

chmod +x ./select_gpu

CPU_BIND="map_cpu:49,57,17,25,1,9,33,41"

export MPICH_GPU_SUPPORT_ENABLED=1

srun --unbuffered --cpu-bind=${CPU_BIND} ./select_gpu ./neko rugby_ball.case
rm -rf ./select_gpu
```

gloopydoop: > And do you also want a LUMI-small for the small partition?

No, that would be overkill

timfelle: So you can never submit a small job?

gloopydoop: I guess the point is, in Nek5000 and neko, I think everyone is responsible for coming up with their own job scripts or passing them around etc. We either go in the direction of them, and potentially set up dry runs and leave it to the actual job submission to the user. Or we go in this direction, where we provide job scripts (or at least templates)

And what I'm saying is that provide a jobscript with just the name `LUMI` is misleading. The big distinction in my mind is between lumi-c and lumi-g since they will have radically different job scripts. Regarding lumi-small, that can probably be handled with a well executed sed command :sweat_smile:

timfelle: Ahh I see the issue now.

No, there is not a single jobscript called LUMI. There is a folder called LUMI where a user must place a jobscript with specific instructions from the specific example they wish to execute. This allow a user to keep different scripts for different clusters in the same structure.

gloopydoop: tip top gumdrop,
This is all clear :+1:

gloopydoop (approved):
yehaw :cowboy_hat_face:

PR #173, opened 2025-04-24
Set default projection space size to 0 for solvers

timfelle:
Update the default projection space size to 0 in the configuration for both velocity and pressure solvers, simplifying the examples.


gloopydoop (approved):
nice!
I was lurking a little on zulip :stuck_out_tongue_winking_eye:

PR #175, opened 2025-04-24
Fix device pointer issue pointed out by Harry.

timfelle:
The assignment operator does not do a deep enough copy for the objectives and constraints.


gloopydoop (approved):
tip top :+1:

PR #162, opened 2025-04-09
Extended neko-top to support HIP

gloopydoop:
This PR takes a lot of our current CUDA wrappers and kernels and adds HIP counterparts.

***note*** I'm still wrestling a lot with LUMI so this is on draft until #107 is complete.

PR #177, opened 2025-04-25
avoid masked kernel launches when no GLL pts are in the mask

gloopydoop:
This is small PR that was found as part of #162 but should be corrected for cuda first.

Essentially, if your partition contains no elements inside the mask, the you launch a kernel of zero size. This PR avoids this edge case.


sources/neko_ext/math/bcknd/device/cuda/math_ext.cu
```diff
@@ -51,6 +51,7 @@ void cuda_copy_mask(void* a, void* b, int* size, int* mask, int* mask_size) {
     const dim3 nthrds(1024, 1, 1);
     const dim3 nblcks(((*mask_size) + 1024 - 1) / 1024, 1, 1);
 
+    if(*mask_size == 0) return;
```
timfelle:
```suggestion
    if (*mask_size == 0) return;
```

timfelle (approved):

timfelle (approved):

PR #178, opened 2025-04-29
Gradient jump is now a source term

timfelle:


gloopydoop (approved):
nice work staying vigilant on the `neko` PRs :+1:

PR #181, opened 2025-04-30
`--fresh` and `--clean-first` does not work for fortran. Resorting to complete removal of build directory instead.

timfelle:


gloopydoop (approved):
I've actually been doing this manually on occasion, so this is good :+1: 

Correct me if I'm wrong, because this is how I'm interpreting these changes, if I go:
`./setup.sh -c --clean-neko` I can be COMPLETELY certain that this is a complete rebuild. Particularly relevant when compiling on device. ie, `./setup.sh -c --clean-neko -d CUDA` should ensure there is no chance of left over files which have not been compiled for device. Is this correct?

PR #183, opened 2025-05-01
Fix bug from latest neko update

timfelle:
Resolve an issue with logging time-step status by updating the method call to ensure correct functionality.


gloopydoop (approved):
👍

AbbasMoosaV (approved):

PR #184, opened 2025-05-06
Update MMA functions and ensure array compatibility

timfelle:
Refactor MMA function interfaces for improved clarity and ensure that arrays are correctly initialized and updated based on device support. This enhances the overall functionality and performance of the MMA module.

Pure CPU and pure GPU works. Mixing does not, so the default is currently what works.

Let's get this in to make


AbbasMoosaV (approved):
I checked and this works perfectly  for example/mma test case for all cases.
Even when neko_bcknd is set to cuda and we use cpu bcknd in mma, it still works for both dip and dpip.
Therefore, I approve this and merge it in.

PR #187, opened 2025-05-07
Resolve issue in mixed backend

timfelle:
Adjustments made to ensure proper handling of input and output vectors in the mixed backend, enhancing memory management and data transfer between host and device.

### DIFF
--- a/examples/rugby_ball/prepare.sh
+++ b/examples/rugby_ball/prepare.sh
@@ -17,16 +17,14 @@ function help() {
     echo -e ""
     echo -e " Options:"
     echo -e "  -h, --help  Show this help message and exit."
-    echo -e "  -x#         Number of cells in the x direction."
-    echo -e "  -y#         Number of cells in the y direction."
-    echo -e "  -z#         Number of cells in the z direction."
+    echo -e "  -N#         Number of cells in the x and y direction."
     echo -e ""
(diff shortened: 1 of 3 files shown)


examples/rugby_ball/prepare.sh
```diff
@@ -61,8 +58,12 @@ fi
 # ============================================================================ #
 # Generate mesh and run case
 
+# Compute the depth to retain the aspect ratio of the elements
+# The aspect ratio each element is  1:1:1
+z=$(python3 -c "print(1.0 / $N)")
+
 echo "Generating mesh with dimensions: $Nx $Ny $Nz"
-genmeshbox 0 1 0 1 0 1 $Nx $Ny $Nz .false. .true. .true.
+genmeshbox 0 1 0 1 0 $z $Nx $Ny $Nz .false. .true. .true.
```
gloopydoop, line 66: should we consider the same modification for `rugby_veriification` ?
I guess it's not super important since that will become a unit test soon

gloopydoop (approved):
Nice work :+1:

timfelle (commented):
> should we consider the same modification for `rugby_veriification` ?
> I guess it's not super important since that will become a unit test soon 

Lets do that later when we work on that example anyway

PR #188, opened 2025-05-08
Refactor user material properties handling

timfelle:
Realign with neko.


gloopydoop (approved):
Great work mate :+1: 
Compiled, adjoint ran, we can sort out the finer details in a different PR.

Coming through in a clutch :saluting_face:

PR #193, opened 2025-05-10
Reset adjoint fluid and adjoint scalar

gloopydoop:
The PR adds functionality to reset the adjoint fluid and scalar after an optimization iteration.

PR #196, opened 2025-05-12
Mixer example.

gloopydoop:
This PR contains parameters which I used successfully to run the mixer example.
***note:***  `this%mma%get_residumax()` appears to be broken, so it's been commented out. This should be fixed before pulling in this PR, and this PR should be updated.

Depends on 
#194 
#193 
#191
an upcoming PR on postprocessing on the branch docs/mixer_post_processing which is still a work in progress.


sources/optimizer/mma_optimizer.f90
```diff
@@ -175,7 +175,7 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
     call design%write(0)
 
     do iter = 1, this%max_iterations
-       if (this%mma%get_residumax() .lt. this%tolerance) exit
+       ! if (this%mma%get_residumax() .lt. this%tolerance) exit
```
```suggestion
       if (this%mma%get_residumax() .lt. this%tolerance) exit
```

examples/passive_scalar/user.f90
```diff
@@ -96,6 +96,7 @@ subroutine scalar_ic(s, params)
     ! Initial scalar profile is a sigmoid separating the two species
     do i = 1, s%dof%size()
        s%x(i,1,1,1) = L / (1.0_rp + exp(-k*(s%dof%z(i,1,1,1) - z_0)))
+       ! s%x(i,1,1,1) = 0.0_rp
```
timfelle:
```suggestion
```

PR #197, opened 2025-05-12
Post processing and documentation for the mixer

gloopydoop:
This PR contains
- a python script for plotting the convergence history (it's very simple... maybe we should consider something a bit nicer or removing it completely)
- A `.pvsm` file for post processing the mixer. It looks quite nice I think! I included the university logos because I thought it would look cool, but actually in my opinion it's a bit cluttered and I would vote to remove them and only keep the CEEC logo. But we'll put it to vote in the meeting on Wednesday.
heavy

*** NOTE *** I've put the "help wanted" because maybe it's better to keep the `.pvsm` in some other folder and I'm not so sure how well this will play with `run.sh` on completion.

Maybe it should just be a part of the mixer tutorial.


examples/passive_scalar/post_processing/plot_convergence.py
```diff
@@ -0,0 +1,21 @@
+import pandas as pd
+
+data = pd.read_csv('../optimization_data.csv')
+iter = pd.DataFrame(data['iter']).to_numpy()
+obj = pd.DataFrame(data[' Scalar Mixing']).to_numpy()
+KKT = pd.DataFrame(data[' KKTmax']).to_numpy()
+import matplotlib.pyplot as plt
+plt.rc('text', usetex=True)
+plt.rc('font', family='serif')
+import matplotlib as mpl
```
timfelle:
```suggestion
```

examples/passive_scalar/post_processing/plot_convergence.py
```diff
@@ -0,0 +1,21 @@
+import pandas as pd
+
+data = pd.read_csv('../optimization_data.csv')
+iter = pd.DataFrame(data['iter']).to_numpy()
+obj = pd.DataFrame(data[' Scalar Mixing']).to_numpy()
+KKT = pd.DataFrame(data[' KKTmax']).to_numpy()
+import matplotlib.pyplot as plt
```
timfelle:
```suggestion
```

examples/passive_scalar/post_processing/plot_convergence.py
```diff
@@ -0,0 +1,21 @@
+import pandas as pd
+
```
timfelle:
```suggestion
import pandas as pd
import matplotlib.pyplot as plt

```

timfelle (approved):

gloopydoop (commented):
:+1:

PR #201, opened 2025-05-19
Switch MMA backend keyword from cuda to device

timfelle:
Change the backend keyword for the MMA configuration from "cuda" to "device" to standardize backend references.


gloopydoop (approved):
perfect

PR #202, opened 2025-05-19
Add option for scalar field inclusion in steady simulation component

timfelle:
Introduce a user-selectable option to include or exclude the scalar field in the steady simulation component.

### DIFF
--- a/examples/passive_scalar/passive_scalar.case
+++ b/examples/passive_scalar/passive_scalar.case
@@ -5,7 +5,7 @@
         "output_checkpoints": false,
         "output_at_end": false,
         "output_boundary": false,
-        "end_time": 1e-3,
+        "end_time": 5.0,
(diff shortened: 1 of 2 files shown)


gloopydoop (approved):
This is a really good point, and perfectly applicable to what we're doing RN. So good PR :+1: 

general comment: we're a bit in limbo here, that we're the only people who run steady cases, so I suppose there are some things on the `neko` side we should investigate in upcoming weeks:
- currently scalar has no freeze option
- I still feel like even within `simulation.f90` there should be some kind of exit condition.  not just freezing. Maybe this is accessible by simcomps, maybe by some global variable. Still not sure. Just jotting this down so we don't forget.

Either way, good PR, good thinking, especially in the context on high Prandtl number, like we have in the mixer (right now...)

PR #206, opened 2025-05-26
Enhancement/type_injection

timfelle:
Utilize the new type injection system in Neko.

This solves the user_simcomp bug spotted by harry.


gloopydoop (approved):
:+1:

PR #207, opened 2025-05-27
Integrate genmeshbox into meshing tool and update documentation

timfelle:
Introduce direct usage of the genmeshbox tool for creating box meshes and enhance the documentation to reflect these changes.


gloopydoop (approved):
This looks good, I like the documentation too :+1: 

What's our plan regarding the examples and `prepare.sh`? 
I guess we have to assume there will be a lot of copying and pasting when making a new example, so if we want `mesh.sh` to be a standard tool we should adjusting our current examples for clarity.

PR #212, opened 2025-06-04
Implement backward simulation with adjoint fluid model

timfelle:
Realign the overall structure to neko again. This is a big refactor of the adjoint.

### DIFF
--- a/sources/adjoint/adjoint_fluid_pnpn.f90
+++ b/sources/adjoint/adjoint_fluid_pnpn.f90
@@ -320,14 +320,16 @@ contains
          this%full_stress_formulation, .false.)
 
     if (this%full_stress_formulation .eqv. .true.) then
-       ! Setup backend dependent Ax routines
-       call ax_helm_factory(this%Ax_vel, full_formulation = .true.)
+       call neko_error( &
+            "Full stress formulation is not supported in the adjoint module.")
+       !  ! Setup backend dependent Ax routines
+       !  call ax_helm_factory(this%Ax_vel, full_formulation = .true.)
 
-       ! Setup backend dependent prs residual routines
-       call pnpn_prs_res_stress_factory(this%prs_res)
+       !  ! Setup backend dependent prs residual routines
+       !  call pnpn_prs_res_stress_factory(this%prs_res)
 
-       ! Setup backend dependent vel residual routines
-       call pnpn_vel_res_stress_factory(this%vel_res)
+       !  ! Setup backend dependent vel residual routines
+       !  call pnpn_vel_res_stress_factory(this%vel_res)
     else
        ! Setup backend dependent Ax routines
        call ax_helm_factory(this%Ax_vel, full_formulation = .false.)
--- a/sources/adjoint/adjoint_fluid_pnpn.f90
+++ b/sources/adjoint/adjoint_fluid_pnpn.f90
@@ -669,10 +669,7 @@ contains
 
   !> Advance fluid simulation in time.
   !! @param this The fluid simulation object.
-  !! @param t The time value.
-  !! @param tstep The current interation.
-  !! @param dt The timestep
-  !! @param ext_bdf Time integration logic.
+  !! @param time The time state object.
   !! @param dt_controller timestep controller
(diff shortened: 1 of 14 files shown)


sources/adjoint/adjoint_fluid_pnpn.f90
```diff
@@ -315,7 +312,14 @@ subroutine adjoint_fluid_pnpn_init(this, msh, lx, params, user, time_scheme)
...
 
-    if (this%variable_material_properties .eqv. .true.) then
+    call json_get(params, 'case.numerics.time_order', integer_val)
+    allocate(this%ext_bdf)
+    call this%ext_bdf%init(integer_val)
+
+    call json_get_or_default(params, "case.fluid.full_stress_formulation", &
+         this%full_stress_formulation, .false.)
+
+    if (this%full_stress_formulation .eqv. .true.) then
```
gloopydoop, line 322: hhhmmmm. Maybe we should throw an error here. Have we tested full stress + adjoint?

sources/adjoint/adjoint_fluid_pnpn.f90
```diff
@@ -431,7 +422,7 @@ subroutine adjoint_fluid_pnpn_init(this, msh, lx, params, user, time_scheme)
          'case.fluid.pressure_solver.preconditioner', precon_params)
     call json_get(params, 'case.fluid.pressure_solver.absolute_tolerance', &
          abs_tol)
-    call json_get_or_default(params, 'case.fluid.velocity_solver.monitor', &
+    call json_get_or_default(params, 'case.fluid.pressure_solver.monitor', &
```
gloopydoop, line 427: wow good catch

sources/adjoint/adjoint_fluid_scheme_incompressible.f90
```diff
@@ -0,0 +1,763 @@
...
+       & file!"
+       call neko_log%message(log_buf)
+       this%user_material_properties => user%material_properties
+
+       call user%material_properties(0.0_rp, 0, this%name, &
+            this%material_properties)
+
+    else
+       this%user_material_properties => dummy_user_material_properties
+       ! Incorrect user input
```
gloopydoop, line 703: I can't imagine a time that we would have a separate Re for the adjoint compared to the forward, so I think this is fine to not use the fallback json

sources/adjoint/adjoint_scalar_pnpn.f90
```diff
@@ -481,13 +483,11 @@ subroutine adjoint_scalar_pnpn_setup_bcs_(this, user)
     logical, allocatable :: marked_zones(:)
     integer, allocatable :: zone_indices(:)
 
-
-    if (this%params%valid_path('case.adjoint_scalar.boundary_conditions')) then
-       call this%params%info('case.adjoint_scalar.boundary_conditions', &
+    if (this%params%valid_path('boundary_conditions')) then
```
gloopydoop, line 486: ooooohhhh ok this one is problematic. We need a way to prescribe different BCs on the forward and adjoint.
Maybe we have to include an `adjoint_boundary_conditions` in the `.case`

gloopydoop (commented):
Great work :+1:  Big PR.

I only spotted one thing problematic, and it was how we now handle adjoint scalar BCs, as it appears to me that right now we're forced to have the same BCs for the forward and adjoint scalar. Perhaps another entry in the `.case` file is a solution?

sources/adjoint/adjoint_scalar_pnpn.f90
gloopydoop: ooooooooooohhhh maybe I'm misunderstanding! Are we passing in a subdict here??

timfelle: It is a subdictionary yes

gloopydoop (approved):

sources/adjoint/adjoint_fluid_scheme_incompressible.f90
timfelle: We are not falling back here, so what are you referring to?

PR #214, opened 2025-06-12
Reset time for adjoint case in simulation reset

timfelle:
Ensure that the time and time step for the adjoint case are reset to their initial values during the simulation reset process.


gloopydoop (approved):
looks good for now. I guess you're actively working on this right now, and we haven't touched `simulation.f90` in a while. Upon revisiting I feel like there's a lot of restructuring we could do that would clean things up. But I guess this is what you're working on right now.

PR #231, opened 2025-07-21
Fix volume constraint sensitivity

### LINKED ISSUE #233: Volume constraint

### LINKED ISSUE #248: Sensitivity of volume constraint

timfelle:
Introduce a unit test computing the finite difference of the volume constraint to estimate sensitivities.


sources/problems/constraints/volume_constraint.f90
```diff
@@ -270,39 +274,42 @@ function volume_brinkman_design(this, design) result(volume)
...
     type(brinkman_design_t), intent(in) :: design
     real(kind=rp) :: volume
-    type(field_t), pointer :: work, design_indicator
+    type(field_t), pointer :: work
+    type(vector_t) :: values
     integer :: temp_indices(1)
 
     volume = 0.0_rp
-    design_indicator => neko_field_registry%get_field("design_indicator")
+    values = design%get_values()
```
gloopydoop, line 282: this is very good :+1: glad we're drifting away from the registry

examples/rugby_ball/rugby_ball.case
```diff
@@ -132,15 +132,14 @@
...
+                "scale": 10.0,
                 "auto_scale": false,
                 "xmin": 0.0,
                 "xmax": 1.0,
                 "a0": 1.0,
                 "a": 0.0,
-                "c": 100.0,
-                "d": 0.0
+                "c": 1000.0,
+                "d": 1.0
```
gloopydoop, line 142: is this change necessary for this PR? I'm not opposed to it...

tests/unit/sensitivity/sensitivity.f90
```diff
@@ -0,0 +1,118 @@
...
+    write(*, fmt_head) "Perturbation", "Constraint", "FD Estimate", "Error"
+    write(*, fmt_data) 0.0_rp, constraint, target_sensitivities%x(i), 0.0_rp
+
+    n_perturbations = size(perturbations)
+    do ip = 1, n_perturbations
+       perturb = perturbations(ip)
+       tol = perturb / maxval(perturbations) * tolerance
+
+       ! Ensure the perturbation stays within the bounds of the design variable
+       if (design_vector%x(i) .gt. 0.5_rp) perturb = -perturb
```
gloopydoop, line 64: perhaps a little hard coded with the `0.5rp`, but I still think it should be fine since we assume $\rho \in [0,1]$.

I can't put my finger on this exactly, but I think I think a volume constraint (being linear) is a bit too kind. When we start comparing FD vs adjoint to more complex objectives I think we need to be very careful of how we define the FD. (forward difference, backward difference, central difference etc) To my understanding, what is happening with this switch between the `perturb = -perturb` in certain cases is we're switching between forward and backwards differencing. Again, I think for a linear constraint this should be ok (and potentially ok in all circumstance, as I said, I can't put my finger on why this feels off)

gloopydoop (commented):
This PR is NICE!
It looks like it can be easily extended to testing sensitivity of our objective functions too which is really nice functionality, looking forward to that.

I haven't been able to test this yet because `neko` #2026 made `vector_t%n` private which breaks some of `mask_ops.f90`, but I can see that gets fixed in #293. So I'll come back to this after I've checked out #293 and tested this.

But overall very nice PR!!!

tests/unit/sensitivity/sensitivity.f90
timfelle: yes, the 0.5 is from the volume constriant explicitly, and was not abstracted in a good way it seems.

The perturbation sign change should not switch from a forward to backward difference. Forward and 

Forward: df_f = (f(x + dx) - f(x)) / dx
Backward: df_b = (f(x) - f(x - dx)) / dx

examples/rugby_ball/rugby_ball.case
timfelle: If we want the example to converge to a non-zero solution then yes

timfelle (commented):
@gloopydoop Should we get this into the Sensitivity branch. It would make it a lot simpler for the other FD checks to be based off the same point

PR #240, opened 2025-07-30
Introduce a validation function for optimizers

timfelle:


sources/optimizer/mma_optimizer.f90
```diff
@@ -235,6 +238,31 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
...
+    call problem%get_constraint_values(constraint_values)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(constraint_values%x, constraint_values%x_d, &
+            constraint_values%size(), HOST_TO_DEVICE, .true.)
+    end if
+
+    if (any(constraint_values%x .gt. 0.0_rp)) then
+       call neko_error("MMA optimizer validation failed: " // &
+            "Constraints are not satisfied.")
+    end if
```
gloopydoop, line 259: I'm interpreting this to mean we must start at a feasible point.
It was my understanding this is not a strict requirement of MMA since constraints are ultimately penalties. But perhaps this is a PR for @AbbasMoosaV to review.

AbbasMoosaV: I think its OK to check if constraints are satisfied or not, but this is not a validation.
What MMA actually minimizes is:
` f_0(x) + sum(c_i*y_i + 0.5*d_i*(y_i)^2)`
where the values for y_i, somehow represents whether to the primary constraints are satisfied.
But this depends on the values of different parameters d_i, c_i, ....
If you want to make sure that constraints are always feasible, use large enough values for d_i and c_i.
But then the objective function would be ignored and MMA only ensures the feasibility wrt constraints.
I suggest we call this **primary constraint feasibility check** instead of **validation**.

timfelle: > I'm interpreting this to mean we must start at a feasible point. It was my understanding this is not a strict requirement of MMA since constraints are ultimately penalties. But perhaps this is a PR for @AbbasMoosaV to review.

The function is called after the optimization is done. So it tells you if your final solution is feasible.

AbbasMoosaV (approved):

PR #241, opened 2025-07-30
Fix matrix handling and update backend support

timfelle:
Correct the matrix configuration for device types and enhance backend handling in the build scripts. Add support for CUDA runtime and streamline device type checks.


gloopydoop (approved):
:+1:  looks good!!

PR #243, opened 2025-08-04
sensitivity getter functions

gloopydoop:
This PR highlights that the chain rule of the mapping cascade being applied at the end of `call problem%compute_sensitivity(design, simulation)` is not the sensitivity entering MMA.

This is perhaps not the most elegant solution, it just highlights the issue.


timfelle (commented):
Please bring this up to date with develop

sources/designs/design_brinkman.f90
```diff
@@ -458,6 +460,20 @@ function brinkman_design_get_design(this) result(values)
...
+    integer :: n
+
+    n = this%size()
+    call values%init(n)
+    call copy(values%x, this%sensitivity%x, n)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_copy(values%x_d, this%sensitivity%x_d, n)
+    end if
+
+  end function brinkman_design_get_sensitivity
```
timfelle: This will in best case cause temporary arrays, in the worst memory leaks. Please
look at the updates on develop to update the function

timfelle (approved):

timfelle (approved):

timfelle (changes requested):
Please have a look at the rugby example with this update. It no longer satisfies the constraints.

timfelle (approved):

PR #250, opened 2025-08-05
Add an augmented lagrangian objective

### LINKED ISSUE #251: Handling of u * u_adjoint (augmented Lagrangian)

timfelle:
Split the adjoint stuff out of minimum dissipation, and create a separate objective

Closes PR #251

### DIFF
--- a/sources/problems/objectives/augmented_lagrangian_objective.f90
+++ b/sources/problems/objectives/augmented_lagrangian_objective.f90
@@ -30,7 +30,7 @@
 ! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 ! POSSIBILITY OF SUCH DAMAGE.
 !
-!> Implements the `augmented_lagrangian_t` type.
+!> Implements the `augmented_lagrangian_objective_t` type.
 !
 ! I promise I'll write this document properly in the future...
 !
@@ -60,7 +60,7 @@
 ! This has always annoyed me...
 ! because now I see one objective and one constraint
 !
-module augmented_lagrangian
+module augmented_lagrangian_objective
   use num_types, only: rp
   use field, only: field_t
   use field_math, only: field_col3, field_addcol3, field_cmult, field_add2s2, &
@@ -68,8 +68,6 @@ module augmented_lagrangian
   use operators, only: grad
   use adjoint_fluid_pnpn, only: adjoint_fluid_pnpn_t
   use scratch_registry, only: neko_scratch_registry
-  use adjoint_augmented_lagrangian_source_term, only: &
-       adjoint_augmented_lagrangian_source_term_t
   use objective, only: objective_t
   use simulation_m, only: simulation_t
   use adjoint_fluid_scheme, only: adjoint_fluid_scheme_t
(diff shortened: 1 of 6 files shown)


sources/problems/objectives/minimum_dissipation_objective.f90
```diff
@@ -284,25 +284,6 @@ end subroutine minimum_dissipation_update_value
   subroutine minimum_dissipation_update_sensitivity(this, design)
     class(minimum_dissipation_objective_t), intent(inout) :: this
     class(design_t), intent(in) :: design
```
gloopydoop, line 286: I think problem still cycles through and collects `this%sensitivity`, so perhaps an `rzero` here.

sources/problems/objectives/augmented_lagrangian_objective.f90
```diff
@@ -0,0 +1,232 @@
...
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Implements the `augmented_lagrangian_objective_t` type.
+!
+! I promise I'll write this document properly in the future...
```
gloopydoop: we can clean all of this up

sources/problems/objectives/augmented_lagrangian_objective.f90
```diff
@@ -0,0 +1,232 @@
...
+  use mask_ops, only: mask_exterior_const
+  use math_ext, only: glsc2_mask
+  use utils, only: neko_error
+  use json_module, only: json_file
+  use json_utils, only: json_get_or_default
+  implicit none
+  private
+
+  !> An objective function corresponding to minimum dissipation
+  !! \f$ F =  \int_\Omega |\nabla u|^2 d \Omega + K \int_Omega \frac{1}{2} \chi
```
gloopydoop: here aswell

gloopydoop (changes requested):
This makes sense. 

I think on the road to the super PR we should also add some documentation, I'm happy to add a commit of documentation in either the theory or user guide in relation to what we discussed and describing what this objective does.

Also, similar to how you deleted the sensitivity in minimum dissipation, we need to do the same for scalar mixing

sources/problems/objectives/minimum_dissipation_objective.f90
timfelle: The array is initialized to 0 when the object is created so there is no need to explicitly assign it to 0 again every time.

gloopydoop: tip top :+1:

gloopydoop (approved):
:+1:

PR #264, opened 2025-09-02
Enhance the checkpointing structure

timfelle:
The checkpointing system is now a seperate module with a dedicated class.

New algorithms can be implemented as submodules and called from the main module as is done with the linear one at the moment.

### DIFF
--- a/sources/simulation/simulation.f90
+++ b/sources/simulation/simulation.f90
@@ -195,7 +195,7 @@ contains
     call json_get_or_default(parameters, "checkpoints.enable", &
          this%checkpoint_enable, .false.)
     if (this%checkpoint_enable) then
-       call this%checkpoint%init(parameters, this%neko_case)
+       call this%checkpoint%init(this%neko_case, parameters)
     end if
 
   end subroutine simulation_initialize
@@ -230,7 +230,7 @@ contains
        call simulation_step(this%neko_case, dt_controller, loop_start)
 
        if (this%checkpoint_enable) then
-          call this%checkpoint%save(this%neko_case, this%neko_case%time)
+          call this%checkpoint%save(this%neko_case)
        end if
     end do
     call profiler_stop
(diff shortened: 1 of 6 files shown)


sources/simulation/simulation.f90
```diff
@@ -306,9 +225,13 @@ subroutine simulation_run_forward(this)
     call profiler_start
     loop_start = MPI_WTIME()
     do while (this%neko_case%time%t .lt. this%neko_case%time%end_time)
```
gloopydoop, line 227: So far I'm really liking this PR!

This is exactly what we need, and unsteady sim and different options for handling the checkpointing. 👍 

This, and line 258, is something that I addressed in an adhoc way in #273 that I think would be better addressed in this PR.

It's my assumption that the reason this while loop is based on time <  end time is to allow people to use variable timesteps in the their simulation, so there won't be a predetermined fixed number of iterations for the simulation. With unsteady adjoints, variable timestepping is not really an option. It's super important we have the exact same number of timesteps in the forward as the adjoint. Also if we choose to implement REVOLVE, the total number of iterations required is an important parameter prescribed apriori.

So the bug I saw currently on develop was in the adjoint and regards line 258. `this%n_timesteps` is not the total number of timesteps, so we never actually step our adjoint backward.

As I said, my solution in #273 was adhock, so if you can address it in this PR, ensuring the forward and adjoint will always be syncronized, that would be great 👍  (otherwise we do a separate PR)

I'm at home now so I haven't pulled this in a tested, but it's looking good 👍 I'll finish the review tomorrow

timfelle: This Pr is regarding the saving and restoring of checkpoints and have nothing to do with the simulation solution flow.

Lets work on this in a different PR where it would be relevant. I will condense you comment to an issue for now.

gloopydoop (approved):
looks good :+1:

PR #265, opened 2025-09-03
Introduce an additional coef of the GL space

gloopydoop:
This additional coef on the GL space is used for over integration.
This rather hacky right now needing to create a brand new dofmap, gs etc just to get the coefficients.
I intend to modify things on the neko side allowing a coef to be generated completely with a preexisting gs etc and a new space.

### DIFF
--- a/sources/adjoint/adjoint_fluid_scheme.f90
+++ b/sources/adjoint/adjoint_fluid_scheme.f90
@@ -80,6 +80,7 @@ module adjoint_fluid_scheme
   use field_list, only : field_list_t
   use time_state, only: time_state_t
   use field_math, only: field_addcol3
+  use interpolation, only: interpolator_t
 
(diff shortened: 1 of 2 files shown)


sources/adjoint/adjoint_fluid_scheme.f90
```diff
@@ -99,6 +100,17 @@ module adjoint_fluid_scheme
      type(gs_t) :: gs_Xh !< Gather-scatter associated with \f$ X_h \f$
      type(coef_t) :: c_Xh !< Coefficients associated with \f$ X_h \f$
 
+     ! Tim. This will need to be refactored, but since we have so many extra
```
timfelle, line 103: What need to be refactored about this, and why did you not do so then?

timfelle (approved):

PR #269, opened 2025-09-03
add the potential for a mapping cascade on the volume constraint

gloopydoop:
This allows a used to, for instance, compute the volume constraint based on the filtered volume.
It also allows the PDE filter to be tested in isolation in unit tests

depends on #268

### DIFF
--- a/sources/problems/constraints/volume_constraint.f90
+++ b/sources/problems/constraints/volume_constraint.f90
@@ -170,9 +170,6 @@ contains
     this%limit = limit
     this%c_Xh => simulation%neko_case%fluid%c_Xh
 
-    ! Assume no mapping, unless appended later
-    this%if_mapping = .false.
-
     ! Now we can extract the mask/has_mask from the design
     if (this%has_mask) then
 
--- a/sources/problems/constraints/volume_constraint.f90
+++ b/sources/problems/constraints/volume_constraint.f90
@@ -323,7 +323,7 @@ contains
 
     volume = 0.0_rp
     if (this%if_mapping) then
-       unmapped_values = design%get_values()
+       call design%get_values(unmapped_values)
        call values%init(unmapped_values%size())
        call this%mapping%apply_forward(values, unmapped_values)
        call unmapped_values%free()
--- a/sources/problems/constraints/volume_constraint.f90
+++ b/sources/problems/constraints/volume_constraint.f90
@@ -325,7 +325,7 @@ contains
        call this%mapping%apply_forward(values, unmapped_values)
        call unmapped_values%free()
     else
-       values = design%get_values()
+       call design%get_values(values)
     end if
 
     if (this%has_mask) then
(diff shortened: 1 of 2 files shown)


sources/problems/constraints/volume_constraint.f90
```diff
@@ -73,9 +73,12 @@ module volume_constraint
      real(kind=rp) :: limit
      !> Volume of the optimization domain
      real(kind=rp) :: volume_domain
-
      !> Pointer to the SEM field.
      class(coef_t), pointer :: c_Xh => null()
+     !> Mapping cascade
+     type(mapping_handler_t) :: mapping
+     !> if mapping is needed
+     logical :: if_mapping
```
timfelle: Set the default false here

sources/problems/constraints/volume_constraint.f90
```diff
@@ -155,6 +169,9 @@ subroutine volume_constraint_init_from_components(this, design, simulation, &
     this%limit = limit
     this%c_Xh => simulation%neko_case%fluid%c_Xh
 
+    ! Assume no mapping, unless appended later
+    this%if_mapping = .false.
+
```
timfelle: Do this in the class definition instead
```suggestion
```

sources/problems/constraints/volume_constraint.f90
```diff
@@ -188,14 +205,18 @@ subroutine volume_constraint_init_from_components(this, design, simulation, &
...
        call cmult(this%sensitivity%x, -1.0_rp / this%volume_domain, &
             design%size())
-    end if
 
-    if (this%is_max) then
-       call vector_cmult(this%sensitivity, -1.0_rp)
+       if (this%is_max) then
+          call cmult(this%sensitivity%x, -1.0_rp, design%size())
+       end if
     end if
```
timfelle: Why these changes?

Also, the function `cmult2` would merge the copy and cmult into a single
operation (which would also eliminate a kernel launch.)

sources/problems/constraints/volume_constraint.f90
```diff
@@ -235,8 +256,33 @@ end subroutine volume_constraint_update_value
...
+    type(field_t), pointer :: unmapped, mapped
+    integer :: temp_indices(2)
+
+    if (this%if_mapping) then
+       ! Recompute and map backward
+       call neko_scratch_registry%request_field(unmapped, temp_indices(1))
+       call neko_scratch_registry%request_field(mapped, temp_indices(2))
+       ! The mapping will handle the mass matrix
+       call field_rone(unmapped)
+       call field_cmult(unmapped, -1.0_rp / this%volume_domain)
```
timfelle: You can use cfill instead of applying ones and then multiply by a constant.

timfelle (changes requested):

sources/problems/constraints/volume_constraint.f90
gloopydoop: yeh good point, I think this was a bad merge from an old branch

timfelle (approved):

PR #272, opened 2025-09-03
Update the sensitivity driver

gloopydoop:
This was a big change from the original but the motivation was the following: I wanted to test not just the objectives/constraints, but the whole problem interface too. This lead to the switch from prescribing various `.f90` files to various `.case` files.

Another big change is that Niels and Casper really want to see that linear slope, not just sensitivity within some tolerance. So now we dump a `.csv` file for each test, and they can be plotted with `plotter_FD_check.py`.

I've put the help wanted flag on because I'm incredibly sus on the CMake script, this is not my forte.

Another big consideration is that right now we can only do steady runs, which means we have to wait a LONG time for these tests to complete. However, Casper mention that if we're just interested in verifying the adjoint, we can do that with an unsteady case and only a few timesteps, which is a really good idea (we just don't have that functionality yet) So ideally in the future these tests will become a lot faster.

depends on #243


tests/unit/sensitivity/CMakeLists.txt
```diff
@@ -1,35 +1,51 @@
 # Define the CTest for sensitivity
-
-# Set the CMake folder for organization
 set(CMAKE_FOLDER "Tests/Unit/Sensitivity")
 
-# Define the list of sensitivity tests
-# each of these should be a custom driver.
+# Provide a list of case files instead of .f90 files
```
timfelle: Please document the code as it is. not how it was changed. Nobody can see what
is was before anyway.

tests/unit/sensitivity/CMakeLists.txt
```diff
@@ -1,35 +1,51 @@
...
 add_library(sensitivity_utils EXCLUDE_FROM_ALL
-    ${CMAKE_CURRENT_SOURCE_DIR}/sensitivity.f90)
+    ${CMAKE_CURRENT_SOURCE_DIR}/sensitivity.f90
+    ${CMAKE_CURRENT_SOURCE_DIR}/user.f90
+)
 target_include_directories(sensitivity_utils PUBLIC ${NEKO_TOP_MODDIR})
 target_link_libraries(sensitivity_utils
     PUBLIC
-        Neko-TOP::neko-top
+        neko-top
```
timfelle: That is wrong. You must have the Neko-Top namespace there.

tests/unit/sensitivity/problem_tester.f90
```diff
@@ -69,39 +77,64 @@ program volume_sensitivity
...
+
   if (NEKO_BCKND_DEVICE .eq. 1) then
-     call device_memcpy(sensitivities%x, sensitivities%x_d, &
-          sensitivities%size(), DEVICE_TO_HOST, .true.)
+     call device_memcpy(sensitivities%x, &
+          sensitivities%x_d, sensitivities%size(), &
+          DEVICE_TO_HOST, .true.)
   end if
 
+  ! BE SO CAREFUL THIS DOESN*T WORK IN PARALLEL !
```
timfelle: The tests are not executed with mpi
```suggestion
```

tests/unit/sensitivity/problem_tester.f90
```diff
@@ -35,18 +36,25 @@ program volume_sensitivity
...
+  type(problem_t) :: prob
 
   ! Test specific variables
   real(kind=rp) :: tolerance = 1e-5_rp
-  real(kind=rp), parameter :: perturbations(4) = [ &
-       1e-1_rp, 1e-2_rp, 1e-3_rp, 1e-4_rp]
+!   real(kind=rp), parameter :: perturbations(8) = [ &
+!        1e-1_rp, 1e-2_rp, 1e-3_rp, 1e-4_rp, 1e-5_rp, 1e-6_rp, 1e-7_rp, 1e-8_rp]
+  real(kind=rp), parameter :: perturbations(8) = [ &
+       5e-1_rp, 1e-1_rp, 5e-2_rp, 1e-2_rp, 5e-3_rp, 1e-3_rp, 5e-4_rp, 1e-4_rp]
```
timfelle: why are we only testing to 1e-4?

gloopydoop: This is a good thing to discuss.
- From a "make sure the adjoint is correct" perspective, we were looking for a linear slope and didn't care about the roundoff side.
- From a "useful unit test" perspective, this is not a useful unit test at all (except for the volume and filtered volume, but not for anything that involves a simulation), they take too long to converge steady states. When we get the unsteady stuff off the ground we can do a handful of timesteps and it will execute in a reasonable time.

If Casper is around ask him what he thinks the right thing to do is.

timfelle: Sure, but please modify the perturbation list instead of commenting it out and adding a new line bellow it. Git will show the old stuff and we can decide to modify it or not before merging.

timfelle: Another note is that right now the test attempts to compute the sensitivity of a point with 0 sensitivity, which is not useful.

And the volume sensitivity is not related to the simulation, therefore that test should still be valid.

timfelle: But I guess a solution would be to have this as a regression test running after each update to develop, or nightly, instead of it being a requirement for merger. Unit test should indeed be simple. And the volume constraint should just be a unit test.

gloopydoop: I'm not sure this is the best solution, but specifically this commit: 
https://github.com/ExtremeFLOW/neko-top/pull/272/commits/4aa3c822ea7c9c2ec8d3abb193f41907da99d464

The previous definition of tolerance focuses more on the roundoff side, which is not the intention here.
Still not a perfect solution, ie, not considering the linear region, but for a small test it's ok.

gloopydoop:
The exact purpose of this PR is specifically the driver. It depends on #283 which has been absorbed by this PR, so please review that first.
So far the only objective that gives the perfectly straight line is the lube term. I suspect it's because the forcing for the minimum dissipation is $\nabla \mathbf{u} \cdot \nabla \mathbf{v}$, so at low resolution it's likely that we have discontinuities across element interfaces. This is just a guess... I'll continue looking into it before we finalize #247.

For now, please review just in the context of the driver alone.
image

timfelle (commented):
Is the plan that this should be GPU supported at the moment or do you intend to make that a seperate PR?

gloopydoop:
That will be the LAST PR.
So ideally we fix all the bugs, get perfectly straight lines on CPU, rename those straight lines as the `ref_` versions. And then see if we get a close enough match on GPUs. All of this will still target `sensitivity`.

But this PR is just for the drivers

timfelle (dismissed):

tests/regression/sensitivity/CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
...
+# ./setup.sh -t
+# And then within the build directory run:
+# python plotter_FD_check.py
+# To compare against the reference values
+set(test_list
+    #"minimum_dissipation.case"
+    #"lube_term.case"
+    #"min_dis_and_lube.case"
+    #"min_dis_and_lube_weights.case"
+    #"passive_scalar.case"
```
timfelle: No tests are enabled.

tests/regression/sensitivity/ref_FD_check_lube_term.case.csv
timfelle, line 1: This file have an invalid extension `.case.csv` does not make sense. 
Remove `.case` or convert the `.` to a `_`

tests/regression/sensitivity/CMakeLists.txt
gloopydoop: read the comment above

timfelle: With PR290, regression tests are not executed by the CI for Pr's. So please add uncomment these so that we do not have to manually modify the files to run a test

tests/unit/sensitivity/problem_tester.f90
```diff
@@ -0,0 +1,136 @@
...
+  else if (prob%get_n_constraints() .eq. 1) then
+     ! note we always have a dummy objective
+     is_objective = .false.
+  else
+     write(nobj_str, '(I0)') prob%get_n_objectives()
+     write(ncon_str, '(I0)') prob%get_n_constraints()
+     call neko_error("Specify a) a single constraint b) multiple " // &
+          "objectives. You have" // nobj_str // &
+          "objectives and " // ncon_str // " constraints.")
+  end if
```
timfelle, line 93: This test is for the volume constraints. You know that these are for a
constriant. Please make the test as simple as possible to avoid confusions.

tests/regression/sensitivity/ref_FD_check_lube_term.case.csv
gloopydoop: that's a good point. double check this commit: c12946d047616aa9a32e909e7914937f2a544e65

tests/unit/sensitivity/sensitivity.f90
```diff
@@ -115,4 +116,142 @@ subroutine compute_sensitivity_list(object, sim, des, target_sensitivities, &
     end do
   end subroutine compute_sensitivity_list
 
+  subroutine compute_problem_sensitivity_i(problem, sim, des, target_sensitivities, i, &
+       perturbations, tolerance, file_name, is_objective)
```
timfelle: I do not see this function being used by the driver, so why is it here?

tests/unit/sensitivity/volume.case
```diff
@@ -129,22 +125,14 @@
...
-                "type": "minimum_dissipation",
-                "weight": 1.0
-            },
-            {
-                "type": "lube_term",
-                "weight": 1.0,
-                "K": 1.0
-            }
+            // empty so we don't double up
         ],
```
timfelle: Should this not be removed then?

tests/unit/sensitivity/volume.case
```diff
@@ -129,22 +125,14 @@
...
         "constraints": [
             {
                 "type": "volume",
                 "mask_name": "optimization_domain",
                 "limit": 0.2,
-                "is_max": false
+                "is_max": false,
```
timfelle: Why the extra comma?

tests/unit/sensitivity/volume.case
```diff
@@ -63,7 +63,7 @@
         "simulation_components": [
             {
                 "type": "steady",
-                "tol": 1e-2,
+                "tol": 1e-6,
                 "compute_control": "tsteps",
                 "compute_value": 1
             }
```
timfelle: The code runs a single step. Why do we still have the steady simcomp here?

tests/unit/sensitivity/problem_tester.f90
gloopydoop: for now. 
Ideally we would also test objectives (for a few timesteps) when we have the unsteady stuff up and running.
Right now they're sitting in the regression.

But these were a bit awkward to handle.

tests/regression/sensitivity/problem_tester.f90
timfelle, line 1: The test is invalid. There are no case in which it will return false.
If you intend to define this as a ctest, it must have a non-0 return value to
indicate that the test have failed. Otherwise it is not a test and should just
be a simple executable.

Which would also simplify the cmake file significantly and remove the whole
being run on the CI issue as well.

gloopydoop: yeh, the distinction between unit and regression isn't well handled right now. In our last meeting the suggestion was to just put these tests **somewhere**.
In the future, for an actually regression test, I would envision the test would be a difference between the `FD_check____.csv` and `ref_FD_check____.csv`. But this is going a bit beyond the scope of this PR and my CMake skills

tests/regression/sensitivity/CMakeLists.txt
gloopydoop: I'll take a look

tests/regression/sensitivity/problem_tester.f90
timfelle: But lets not make them a cmake test. I can set that up if you want. It would be very similar to an example, where we just compile the program and then have a littly run script which executes it for each case file and runs the python scripts

gloopydoop: sure, but let's do that in a different PR also targeting #247, because it would also have to be conscious of running in parallel which would require some modification to ´sensitivity.f90´

tests/unit/sensitivity/problem_tester.f90
timfelle: In that case a different driver should be introduced anyway. There is no reason to shoehorn things together which are much simpler when split. Especially for the test stuff. We want those to be as simple and single purpose as possible to eliminate the required work when it triggers.

tests/regression/sensitivity/problem_tester.f90
timfelle: It is much simpler to make a PR targeting this branch with those changes. it would end up being a mess to fix for #247. It is already too big to be able to read through.

timfelle (approved):

PR #273, opened 2025-09-03
Miscellaneous changes

gloopydoop:
These are some misc changes I need to have things running correctly. The big ones are:

- zeroing out lag terms and RHS on reset
- The total number of timesteps in the adjoint going backwards (I actually think we should do the same thing forward, it's important we have the exact same number, and I think the intention of doing a while loop in the forward and looking at the total time is to allow people to use variable timesteps. Which we can use.)

### DIFF
--- a/sources/simulation/simulation.f90
+++ b/sources/simulation/simulation.f90
@@ -101,7 +101,7 @@ module simulation_m
      ! This is used to save the state of the simulation at certain time steps
      ! to be able to restart the simulation from there. This is used for the
      ! adjoint simulation.
-     logical :: checkpoint_enable = .false.
+     logical :: unsteady = .false.
      integer :: n_saves_memory = 10
      integer :: n_saves_disc = 0
      integer :: n_timesteps = 0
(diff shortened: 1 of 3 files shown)


sources/simulation/simulation.f90
```diff
@@ -332,8 +357,9 @@ subroutine simulation_run_backward(this)
     cfl = this%adjoint_case%fluid_adj%compute_cfl(this%adjoint_case%time%dt)
     loop_start = MPI_WTIME()
 
-    do i = this%n_timesteps, 1, -1
-       if (this%checkpoint_enable) call this%restore_state(i)
+    total_timesteps = int(this%neko_case%time%end_time / this%adjoint_case%time%dt)
+    do i = total_timesteps, 1, -1
```
timfelle, line 361: I am quite certain it is better to count the steps we do forward, and then run
back through those steps in the backward, instead of expecting us to run till
the end. 
We might want to add a stopping criterion for the forward based on some
statistics being stable (such as the shedding vortex being stable). If we had
then counted the forward steps, we could run back through them instead of having
a forced end time.

timfelle (changes requested):

gloopydoop:
The PR has been absorbed by #247

sources/simulation/simulation.f90
gloopydoop: Yep I agree, I'm happy with that. My original issue has been resolved indirectly in #279.

PR #276, opened 2025-09-09
Reintroduce support for ParMETIS

timfelle:
Restore functionality to find and install ParMETIS, ensuring it is properly integrated into the build process.

This allows "load_balancing" in case files from Neko, and the "prepart" tool for offline balancing.


gloopydoop (approved):
Built cleanly, no hassle, well done :+1:

PR #282, opened 2025-09-19
Introduce "pressure drop constraint"

gloopydoop:
not exactly a pressure drop constraint. A constraint on the total dissipation. There is a bit of documentation explaining the rational.


gloopydoop:
superseded by #314

PR #284, opened 2025-09-22
Example/1d_beam

AbbasMoosaV:
Update the optimizer and provide various features such as support for performance flag(avoid I/O for performance check) and solving unconstrained problems. Also, add 2 test problems for the MMA implementation: 1- A `1d_beam` example for optimization problems with many constraints 2-`mma` example for unconstrained optimization problem.


examples/1d_beam/driver.f90
```diff
@@ -0,0 +1,334 @@
...
+
+  ! initialize the problem
+  call prob%init(parameters, des)
+
+  allocate(deflection)
+  allocate(beamweight)
+
+  ! Set up distributed stress constraints
+  num_constraints = 10
+  num_constraint_partitions=10
```
gloopydoop: maybe have these as parameters closer to the top of this file. alternatively read from the json?

examples/1d_beam/driver.f90
```diff
@@ -0,0 +1,334 @@
...
+  end do
+
+  ! Add objectives to the problem
+  call prob%add_objective(deflection)
+  call prob%add_objective(beamweight)
+
+
+  call MPI_Barrier(neko_comm, ierr)
+  if (pe_rank == 0) print *, "Constraints distribution complete!"
+  if (pe_rank == 0) print *, "number of problem objectives=", &
```
gloopydoop: print is very awkward, and the current MMA has a ton of them. Try the neko logging so it ends up in the log files properly.

examples/1d_beam/example_problem.f90
```diff
@@ -0,0 +1,469 @@
...
+    ! Exclusive prefix sum for global offset
+    call MPI_Exscan(n, offset, 1, MPI_INTEGER, MPI_SUM, neko_comm, ierr)
+    if (pe_rank == 0) offset = 0
+
+    ! Build elementwise Δ_k on host (geometry-related, not design-related)
+    allocate(Delta(n))
+    do k = 1, n
+       Delta(k) = ((L_total - Le*real(offset+k-1,rp))**3 - &
+            (L_total - Le*real(offset+k,rp))**3) / 3.0_rp
+    end do
```
gloopydoop: a bit nit picky, but surely this can be done on device too?
Especially since this is effectively the "compute" (if I'm understanding correctly)
So we're effectively computing on CPU and copying over the GPU at every iteration.

Maybe be worth a kernel

gloopydoop: (especially because all the vector math already works on GPU, so this is the only bottleneck)

examples/1d_beam/example_problem.f90
```diff
@@ -0,0 +1,469 @@
...
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(h%x, h%x_d, n, DEVICE_TO_HOST, sync = .false.)
+    end if
+
+    ! Compute sensitivity:
+    ! dg/dx = P * Δ * (-36.0 / (E * b * h^4)) * (h_max - h_min)
+    do k = 1, n
+       sensitivity%x(k) = P * Delta(k) * (-36.0_rp) / &
+            (E * b * h%x(k)**4) * (h_max - h_min)
+    end do
```
gloopydoop: same here

examples/1d_beam/readme.md
```diff
@@ -0,0 +1,167 @@
+# A Simple 1D Beam Optimization Test Case for MMA
```
gloopydoop: This whole readme passes doxygen when there are some incorrect syntax. That's because it hasn't been tied into the doxygen. You need to include it in `examples/examples.md`

examples/mma/readme.md
```diff
@@ -0,0 +1,115 @@
+# A Simple Optimization Test Case for MMA
```
gloopydoop: add this one to examples/examples.md too

sources/designs/design_simplefield.f90
```diff
@@ -0,0 +1,226 @@
+! Copyright (c) 2024, The Neko Authors
```
gloopydoop, line 1: I'll leave this one for you and Tim. But I'm confused about design_simple vs design_simplified...

sources/problems/constraints/volume_constraint.f90
```diff
@@ -57,6 +57,8 @@ module volume_constraint
   use math_ext, only: glsc2_mask
   use field_math, only: field_rone, field_copy
   use utils, only: neko_error
+  use vector_math, only: vector_add2, vector_cmult
+
```
gloopydoop: what's this for?

gloopydoop (commented):
I'll leave the lion's share of this review to @timfelle , I just took a quick parse.

My main comment that exists throughout and I stopped pointing them out is `print *` which I think we should avoid because it ends up in the wrong logfile. If you want it in the main logfile use the neko loggers.

The other main one was having CPU commands and memcopies in the `compute` of the objectives and sensitivities. I guess this can be pushed off to a future PR but it really goes against our whole philosophy of "keep EVERYTHING on the GPUs". And the kernels themselves don't look too tricky to implement.

But overall it looks nice, I'll leave the rest to @timfelle

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -42,64 +42,104 @@ __global__ void mma_Ljjxinv_kernel(T* __restrict__ Ljjxinv,
...
+  // Cube once (avoiding pow for speed)
+  const T diff_u3 = diff_u * diff_u * diff_u;
+  const T diff_l3 = diff_l * diff_l * diff_l;
+
+  // Compute inverse value safely
+  T denom = 2.0 * pj / diff_u3 + 2.0 * qj / diff_l3;
+  T val = -1.0 / denom;
+
+  // Mask out active primal constraints
+  bool active = (xt == alpha[tj]) || (xt == beta[tj]);
```
timfelle: Please do not do equality comparison of floating point numbers. You should
expect it to always return false.

Do a compare on the absolute difference instead or something

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -140,88 +180,174 @@ __global__ void mma_sub2_kernel(T* __restrict__ low, T* __restrict__ upp,
...
+  beta[tj] = beta_val;
+
+  // Avoid multiple pow calls, compute once
+  const T upp_minus_x = upp_t - xt;
+  const T x_minus_low = xt - low_t;
+  const T upp_minus_x_sq = upp_minus_x * upp_minus_x;
+  const T x_minus_low_sq = x_minus_low * x_minus_low;
+
+  // Small epsilon for numerical stability
+  const T eps = 1e-5;
```
timfelle: Please use floating point numbers for assigning floating point variables

sources/CMakeLists.txt
```diff
@@ -47,7 +49,7 @@ target_sources(neko-top
         ${CMAKE_CURRENT_SOURCE_DIR}/problems/objectives/minimum_dissipation_objective.f90
         ${CMAKE_CURRENT_SOURCE_DIR}/problems/objectives/scalar_mixing_objective_function.f90
         ${CMAKE_CURRENT_SOURCE_DIR}/problems/constraints/volume_constraint.f90
-
+        ${CMAKE_CURRENT_SOURCE_DIR}/problems/constraints/dummy_constraint.f90
         ${CMAKE_CURRENT_SOURCE_DIR}/optimizer/optimizer.f90
```
timfelle, line 71:
```suggestion
        ${CMAKE_CURRENT_SOURCE_DIR}/problems/constraints/dummy_constraint.f90

        ${CMAKE_CURRENT_SOURCE_DIR}/optimizer/optimizer.f90
```

sources/problems/problem.f90
```diff
@@ -308,14 +309,14 @@ subroutine problem_add_objective(this, objective)
        allocate(this%objective_list(1))
     end if
 
-    call move_alloc(objective, this%objective_list(n + 1)%objective)
+    this%objective_list(n + 1)%objective = objective
```
timfelle: Why this change?

sources/optimizer/optimizer_factory.f90
```diff
@@ -59,8 +63,21 @@ module subroutine optimizer_factory(object, parameters, problem, design, &
...
+    end if
+
+
+    if (present(simulation)) then
+       call object%init_from_json(parameters, problem, design, &
+            max_iterations, tolerance, performance, simulation)
+    else
+       call object%init_from_json(parameters, problem, design, &
+            max_iterations, tolerance, performance)
+    end if
```
timfelle: There is no reason to have this if statement
```suggestion
    call object%init_from_json(parameters, problem, design, &
         max_iterations, tolerance, performance, simulation)
```

sources/optimizer/optimizer_factory.f90
```diff
@@ -59,8 +63,21 @@ module subroutine optimizer_factory(object, parameters, problem, design, &
        call neko_type_error("Optimizer", type, KNOWN_TYPES)
     end select
 
-    call object%init_from_json(parameters, problem, design, &
-         max_iterations, tolerance, simulation)
+    !Check if we are solving an unconstrained problem and add a dummy contraint
```
timfelle: Spelling
```suggestion
    !Check if we are solving an unconstrained problem and add a dummy constraint
```

timfelle (commented):
The rugby_ball run into a memory deallocation error when deallocating the objectives. I expect that to be due to the changes to the non-allocation based objectives.

sources/optimizer/mma_optimizer.f90
```diff
@@ -14,6 +14,7 @@ module mma_optimizer
 
   use vector, only: vector_t
   use matrix, only: matrix_t
+  use vector_math, only: vector_add2, vector_cmult
```
gloopydoop: are these used?

sources/optimizer/bcknd/device/cuda/mma_kernel.h
AbbasMoosaV: I agree, but this is a particular case.
When we update the values of x in the dual subsolve, we clamp the values between alpha and beta:
`x[tj] = fmin(fmax(val, alpha_t), beta_t);`

And then we compute the sensitivities w.r.t x.
So x is either explicitly set equal to alpha or beta by the above line or it is a different number.

In the line that you mentioned:
`bool active = (xt == alpha[tj]) || (xt == beta[tj]);`
I just want to check whether x was clipped by the bounds (alpha or beta) rather than computed through calculations. If it was clipped, we should ignore its sensitivity.

In summary, since we are only checking whether x is explicitly assigned to alpha or beta, using == for this floating-point comparison is safe.

What do you guys think about this?

sources/problems/problem.f90
AbbasMoosaV: I was using an array of constraint in my 1d_beam example and that was not working with constraint and objective being allocatable in constraint_add() and objective_add().

But now I modified 1d_beam and mma example so that we don't need to change anything in the problem.f90.

examples/1d_beam/example_problem.f90
AbbasMoosaV: Since we have a counter on the index of the array (k), this cannot be done through the functions implemented in vector_math.
But we can write another kernel for this calculation.

sources/designs/design_simplefield.f90
AbbasMoosaV: simple_design is a design type where you don't need any msh files. You just provide dimensions and number of elements in x,y, and z direction.
Then a structured mesh is generated in the initializer.

But the simplefield_design requires a msh file and it considers a field_t distributed over the mesh.

So the difference is mostly about whether we use the field_t in neko and the GLL points in a mesh file  or we descritize the domain manually.

sources/optimizer/optimizer_factory.f90
AbbasMoosaV: If we remove the present(simulation) check and always pass it to init_from_json, we get segmentation faults.
The simulation is optionally provided to optimizer factory and if it is not provided, we cannot use 
```
object%init_from_json(parameters, problem, design, &
         max_iterations, tolerance, performance, simulation)
```

For instance, for the mma and 1d_beam examples, I tried this without the check for simulation and I got segmentation faults.

AbbasMoosaV:
> The rugby_ball run into a memory deallocation error when deallocating the objectives. I expect that to be due to the changes to the non-allocation based objectives.


That's true, I was using an array of constraint in the 1d_beam example and I had problems with add_constraints for problem_t.
But now I have modified the 1d_beam example to use an allocatable temp array in a loop to assign all constraint.
Therefore, we can use allocatables for objectives and constraints again, and I already made this change in this PR.

**In short:**
Now the rugby_ball example also works fine on this branch :)

sources/optimizer/optimizer_factory.f90
```diff
@@ -59,8 +64,20 @@ module subroutine optimizer_factory(object, parameters, problem, design, &
...
+       call problem%add_constraint(dummy_con)
+    end if
+
+    if (present(simulation)) then
+       call object%init_from_json(parameters, problem, design, &
+            max_iterations, tolerance, performance, simulation)
+    else
+       call object%init_from_json(parameters, problem, design, &
+            max_iterations, tolerance, performance)
+    end if
```
timfelle: There is no need to check the present stuff. It is done by the object
```suggestion
    call object%init_from_json(parameters, problem, design, &
         max_iterations, tolerance, performance, simulation)
```

sources/optimizer/optimizer_factory.f90
```diff
@@ -59,8 +64,20 @@ module subroutine optimizer_factory(object, parameters, problem, design, &
...
     end select
 
-    call object%init_from_json(parameters, problem, design, &
-         max_iterations, tolerance, simulation)
+    !Check if we are solving an unconstrained problem and add a dummy constraint
+    if (problem%get_n_constraints() .eq. 0) then
+       allocate(dummy_constraint_t::dummy_con)
+       call dummy_con%init(parameters, design)
+       call problem%add_constraint(dummy_con)
+    end if
```
timfelle: This should be handled by the problem. Not the optimizer

sources/optimizer/mma_optimizer.f90
```diff
@@ -41,14 +41,17 @@ module mma_optimizer
...
+     !! constraint_value%x=scale*constraint_value%x)
      !! When auto_scale is true, we use an adaptable scale for
-     !! constraint_value%x and constraint_sensitivities%x in every iteration (variable scale factors)
+     !! constraint_value%x and constraint_sensitivities%x
+     !! in every iteration (variable scale factors)
      real(kind=rp) :: scale
      logical :: auto_scale
 
+     ! when true, then we remove IO loggings for optimal performance
+     logical :: performance
```
timfelle: Please flip this to be a `enable_output` or something like that. The word
"performance" does not describe that output is turned off.

sources/optimizer/mma_optimizer.f90
```diff
@@ -67,14 +70,15 @@ module mma_optimizer
...
-       max_iterations, tolerance, simulation)
+       max_iterations, tolerance, performance, simulation)
     class(mma_optimizer_t), intent(inout) :: this
     type(json_file), intent(inout) :: parameters
     class(problem_t), intent(in) :: problem
     class(design_t), intent(in) :: design
     integer, intent(in) :: max_iterations
     real(kind=rp), intent(in) :: tolerance
     type(simulation_t), optional, intent(in) :: simulation
+    logical, intent(in) :: performance
```
timfelle: Why is this not read from the json file instead of being passed down?

sources/optimizer/mma.f90
```diff
@@ -444,7 +456,7 @@ subroutine mma_update_vector(this, iter, x, df0dx, fval, dfdx)
 
     case ("device")
        call mma_update_device(this, iter, x%x_d, df0dx%x_d, fval%x_d, dfdx%x_d)
-       call device_memcpy(x%x, x%x_d, this%n, DEVICE_TO_HOST, sync = .true.)
+       !  call device_memcpy(x%x, x%x_d, this%n, DEVICE_TO_HOST, sync = .true.)
```
timfelle: If this is not needed anymore, please delete it.

sources/optimizer/mma_optimizer.f90
```diff
@@ -85,11 +89,12 @@ subroutine mma_optimizer_init_from_json(this, parameters, problem, design, &
     ! Initialize the logger
     call this%logger%init('optimization_data.csv')
 
+
     ! Write the header
     problem_header = problem%get_log_header()
     optimization_header = 'iter, ' // trim(problem_header) // &
          ', KKTmax, KKTnorm2, scaling factor'
-    call this%logger%set_header(trim(optimization_header))
+    !  call this%logger%set_header(trim(optimization_header))
```
timfelle: Please remove if it is not needed

sources/optimizer/mma_optimizer.f90
```diff
@@ -85,11 +89,12 @@ subroutine mma_optimizer_init_from_json(this, parameters, problem, design, &
     ! Initialize the logger
     call this%logger%init('optimization_data.csv')
 
+
```
timfelle: Why did you add an additional empty line

sources/optimizer/mma_optimizer.f90
```diff
@@ -168,18 +179,22 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
...
+       call this%logger%write(log_data)
+
+       if (present(simulation)) then
+          call simulation%write(0)
+       end if
+    end if
 
-    if (present(simulation)) call simulation%write(0)
-    call design%write(0)
+    if (.not. this%performance) call design%write(0)
```
timfelle: Why is this not inside the if statement right above, instead of having one more
if statement

sources/optimizer/mma_optimizer.f90
```diff
@@ -217,15 +232,22 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
        call this%mma%KKT(x, objective_sensitivities, &
             constraint_value, constraint_sensitivities)
 
-       ! Stamp the i^th iteration
-       call mma_logger_assemble_data(log_data, iter, objective_value, &
-            all_objectives, constraint_value, this%mma%get_residumax(), &
-            this%mma%get_residunorm(), scaling_factor, &
-            problem%get_n_objectives(), problem%get_n_constraints())
-       call this%logger%write(log_data)
```
timfelle, line 276: Why the empty line?

sources/optimizer/mma_optimizer.f90
```diff
@@ -217,15 +232,22 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
...
+               this%mma%get_residunorm(), scaling_factor, &
+               problem%get_n_objectives(), problem%get_n_constraints())
+          call this%logger%write(log_data)
+
+          if (present(simulation)) then
+             call simulation%write(iter)
+          end if
+       end if
+
+       if (.not. this%performance) call design%write(iter)
```
timfelle: Please merge with the if statement above

sources/optimizer/mma.f90
```diff
@@ -41,7 +41,9 @@ module mma
   use mpi_f08, only: MPI_Allreduce, MPI_INTEGER, MPI_SUM, MPI_COMM_WORLD
   use comm, only: pe_rank
   use utils, only: neko_error
-  use neko_config, only: NEKO_BCKND_DEVICE, NEKO_BCKND_CUDA
+  use neko_config, only: NEKO_BCKND_DEVICE, NEKO_BCKND_CUDA, NEKO_BCKND_HIP, &
+       NEKO_BCKND_OPENCL, NEKO_BCKND_SX, NEKO_BCKND_XSMM
```
timfelle: You never use the SX and XSMM values

sources/optimizer/bcknd/device/device_mma_math.f90
```diff
@@ -62,10 +62,32 @@ module device_mma_math
...
 contains
 
+  subroutine device_delta_1dbeam(Delta_d, L_total, Le, offset, n)
+    !--------------------------------------------------------------------------!
+    ! A device support to do the following calculation for 1D beam elements:   !
+    !   Delta(k) = ((L_total - Le*(offset+k-1))**3 - &                         !
+    !              (L_total - Le*(offset+k))**3) / 3.0_rp                      !
+    !                                                                          !
+    ! Where k ranges from 1 to n                                               !
+    !--------------------------------------------------------------------------!
```
timfelle: Please use doxygen comments.

```suggestion

  !> A device support to do the following calculation for 1D beam elements:
  !! 
  !!   Delta(k) = ((L_total - Le*(offset+k-1))**3 - &
  !!              (L_total - Le*(offset+k))**3) / 3.0_rp
  !!
  !! Where k ranges from 1 to n
  subroutine device_delta_1dbeam(Delta_d, L_total, Le, offset, n)
```

sources/optimizer/bcknd/device/hip/mma_kernel.h
```diff
@@ -36,69 +36,129 @@
...
+__global__ void delta_1dbeam_kernel(T* __restrict__ Delta,
+     const T L_total, const T Le, const int offset, const int n) {
+  int k = blockIdx.x * blockDim.x + threadIdx.x;
+  if (k >= n) return;
+
+  // Convert to 1-based indexing for the calculation
+  int idx = k + 1;
+
+  // Calculate first term: (L_total - Le*(offset+idx-1))^3
+  T x1 = L_total - Le * (T)(offset + idx - 1);
```
timfelle: If you need a cast then please write it explicitly
```suggestion
  T x1 = L_total - Le * static_cast<T>(offset + idx - 1);
```

sources/optimizer/bcknd/device/hip/mma_kernel.h
```diff
@@ -36,69 +36,129 @@
...
+  T term1 = x1 * x1 * x1;
+
+  // Calculate second term: (L_total - Le*(offset+idx))^3  
+  T x2 = L_total - Le * (T)(offset + idx);
```
timfelle:
```suggestion
  T x2 = L_total - Le * static_cast<T>(offset + idx);
```

sources/optimizer/bcknd/device/hip/mma_kernel.h
```diff
@@ -36,69 +36,129 @@
...
+  T term2 = x2 * x2 * x2;
+
+  // Final result
+  Delta[k] = (term1 - term2) / (T)3.0;
```
timfelle:
```suggestion
  Delta[k] = (term1 - term2) / static_cast<T>(3.0);
```

sources/optimizer/bcknd/device/hip/mma_kernel.h
```diff
@@ -36,69 +36,129 @@
...
-    bool is_alpha = xt == alpha[tj];
-    bool is_beta  = xt == beta[tj];
-    Ljjxinv[tj] = (is_alpha || is_beta) ? T(0.0) : val;
-  }
-}
+  if (tj >= n) return;
+
+  // Load inputs into registers
+  const T xt    = x[tj];
+  const T low_t = low[tj];
```
timfelle: Please avoid using `_t` for names as that is commonly used to define type names.
Use the index j instead.

```suggestion
  const T low_j = low[tj];
```

sources/optimizer/bcknd/device/cuda/mma_kernel.h
timfelle: Problem is, it is not safe. This snippit have no guarantee of returning true:
```c++
double x = 0.0;
if (x == 0.0) return true;
return false;
```
Depending on your compiler it might return true and it might return false. The language standard give you no guarantee.

You have to do `abs(xt - alpha[tj]) <= 1e-16` or something similar.

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -35,71 +35,132 @@
...
+__global__ void delta_1dbeam_kernel(T* __restrict__ Delta,
+     const T L_total, const T Le, const int offset, const int n) {
+  int k = blockIdx.x * blockDim.x + threadIdx.x;
+  if (k >= n) return;
+
+  // Convert to 1-based indexing for the calculation
+  int idx = k + 1;
+
+  // Calculate first term: (L_total - Le*(offset+idx-1))^3
+  T x1 = L_total - Le * (T)(offset + idx - 1);
```
timfelle:
```suggestion
  T x1 = L_total - Le * static_cast<T>(offset + idx - 1);
```

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -35,71 +35,132 @@
...
+  T term1 = x1 * x1 * x1;
+
+  // Calculate second term: (L_total - Le*(offset+idx))^3  
+  T x2 = L_total - Le * (T)(offset + idx);
```
timfelle:
```suggestion
  T x2 = L_total - Le * static_cast<T>(offset + idx);
```

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -35,71 +35,132 @@
...
+  T term2 = x2 * x2 * x2;
+
+  // Final result
+  Delta[k] = (term1 - term2) / (T)3.0;
```
timfelle:
```suggestion
  Delta[k] = (term1 - term2) / static_cast<T>(3.0);
```

sources/optimizer/bcknd/device/cuda/mma_kernel.h
timfelle: In fact, you compiler is free to optimize your code to `bool active = false;`

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -35,71 +35,132 @@
...
-    bool is_alpha = xt == alpha[tj];
-    bool is_beta  = xt == beta[tj];
-    Ljjxinv[tj] = (is_alpha || is_beta) ? T(0.0) : val;
-  }
-}
+  if (tj >= n) return;
+
+  // Load inputs into registers
+  const T xt    = x[tj];
+  const T low_t = low[tj];
```
timfelle: Please do not use `_t` as a postfix. It is very commonly used to distinguish
types from variables.
```suggestion
  const T low_j = low[tj];
```

sources/optimizer/bcknd/device/cuda/mma_kernel.h
```diff
@@ -450,38 +668,58 @@ __global__ void mmasumbb_kernel(const T*  __restrict__ GG,
...
+{
+    const int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    const int stride = blockDim.x * gridDim.x;
+
+    // Warp lane and warp id within the block
+    const unsigned int lane = threadIdx.x % warpSize;
+    const unsigned int wid = threadIdx.x / warpSize;
+
+    // Shared memory sized by number of warps per block
+    extern __shared__ T shared[];
```
timfelle: You have to give it a size. Or allocate it. This is a static array, not a
dynamic one.

sources/designs/design_simplefield.f90
```diff
@@ -0,0 +1,226 @@
...
+     private
+
+     type(vector_t) :: values
+     type(vector_t) :: x_coord
+     type(vector_t) :: y_coord
+     type(vector_t) :: z_coord
+
+     ! needed to write the design
+     type(fld_file_output_t), private :: output
+     type(field_t), private :: designfield
```
timfelle: Why do you have both a field and a vector for the values?

timfelle (changes requested):

sources/designs/design_simplefield.f90
AbbasMoosaV: The field type `type(field_t), private :: designfield
` is only used so that we can use the field output function from neko `fld_file_output_t`.
So we only use the `vector_t` in the optimizer.
But, we also add a `field_t` to the design which is updated every time we update the design vector and it is just there to make it easier to write the design to a file by using the fld_file_output_t from Neko.

Otherwise we need to add a specific write function for `vector_t`.
(I think we don't have a `vector_t` file_output in Neko right?)

timfelle: Well then remove the vector variant and just use the field everywhere. There is no reason to keep a permanent copy of the variable for output alone.

sources/optimizer/bcknd/device/device_mma_math.f90
```diff
@@ -83,14 +106,14 @@ end subroutine device_Hess
...
-    ! Ljjxinv = merge(0.0_rp, Ljjxinv, x .eq. beta)                            !
+    !> A device support to do the following calculation needed for the         !
+    !! dualsubsolve for MMA:                                                   !
+    !!   Ljjxinv= - 1 / ( (2*pjlambda/(upp - x)**3) + &                        !
+    !!                                  (2*qjlambda/(x - low)**3))             !
+    !!                                                                         !
+    !! And then remove the sensitivity for the active primal constraints       !
+    !! Ljjxinv = merge(0.0_rp, Ljjxinv, x .eq. alpha)                          !
+    !! Ljjxinv = merge(0.0_rp, Ljjxinv, x .eq. beta)                           !
     !--------------------------------------------------------------------------!
```
timfelle: Documentation of subroutines and functions must be done preceeding the
definition. Otherwise doxygen will not pick it up.

The general structure is:
```
!> Some short function name
!! Longer description, possibly including math and 
!! other things.
subroutine my_procedure()
```

```suggestion
  !> A device support to do the following calculation needed for the
  !! dualsubsolve for MMA:
  !!   Ljjxinv= - 1 / ( (2*pjlambda/(upp - x)**3) + &
  !!                                  (2*qjlambda/(x - low)**3))
  !!
  !! And then remove the sensitivity for the active primal constraints
  !! Ljjxinv = merge(0.0_rp, Ljjxinv, x .eq. alpha)
  !! Ljjxinv = merge(0.0_rp, Ljjxinv, x .eq. beta)
  subroutine device_mma_Ljjxinv(Ljjxinv_d,pjlambda_d, qjlambda_d, x_d, &
       low_d, upp_d, alpha_d, beta_d, n)
```

sources/optimizer/bcknd/device/device_mma_math.f90
```diff
@@ -112,10 +135,10 @@ end subroutine device_mma_Ljjxinv
...
-    ! A device support to do the following calculation needed for the          !
-    ! dualsubsolve for MMA:                                                    !
-    !   x = (sqrt(pjlambda) * low + sqrt(qjlambda) * upp) /  &                 !
-    !                                  (sqrt(pjlambda) + sqrt(qjlambda))       !                                                     !
+    !> A device support to do the following calculation needed for the         !
+    !! dualsubsolve for MMA:                                                   !
+    !!   x = (sqrt(pjlambda) * low + sqrt(qjlambda) * upp) /  &                !
+    !!                                  (sqrt(pjlambda) + sqrt(qjlambda))      !
     !--------------------------------------------------------------------------!
```
timfelle:
```suggestion
  !> A device support to do the following calculation needed for the
  !! dualsubsolve for MMA:
  !!   x = (sqrt(pjlambda) * low + sqrt(qjlambda) * upp) /  &
  !!                                  (sqrt(pjlambda) + sqrt(qjlambda))
  subroutine device_mma_dipsolvesub1(x_d, pjlambda_d, qjlambda_d, &
       low_d, upp_d, alpha_d, beta_d, n)
    
```

sources/optimizer/bcknd/device/device_mma_math.f90
```diff
@@ -136,10 +159,10 @@ end subroutine device_mma_dipsolvesub1
...
-    ! A device support to do the following matrix multiplication               !
-    !               output = matmul(transpose(pij), lambda)                    !
-    ! where matrix pij is mxn, vector lambda is of size m and the output       !
-    ! vector is of size n                                                      !
+    !> A device support to do the following matrix multiplication              !
+    !!               output = matmul(transpose(pij), lambda)                   !
+    !! where matrix pij is mxn, vector lambda is of size m and the output      !
+    !! vector is of size n                                                     !
     !--------------------------------------------------------------------------!
```
timfelle:
```suggestion
  !> A device support to do the following matrix multiplication
  !!               output = matmul(transpose(pij), lambda)
  !! where matrix pij is mxn, vector lambda is of size m and the output
  !! vector is of size n
  subroutine device_mattrans_v_mul(output_d, pij_d, lambda_d, m, n)

```

sources/designs/design_simple.f90
```diff
@@ -253,6 +274,59 @@ subroutine design_simple_write(this, idx)
...
+    ! Scalars
+    write(10,'(A,1X,I0)') 'POINT_DATA', npts
+    write(10,'(A)') 'SCALARS density float 1'
+    write(10,'(A)') 'LOOKUP_TABLE default'
+    do i = 1, npts
+       write(10,'(F20.12)') this%values%x(i)
+    end do
+
+    close(10)
   end subroutine design_simple_write
```
timfelle, line 329: Hard coded file units might be in use by other parts of the code so please don't
use it. Instead use the `newunit` function.
```suggestion
    class(simple_design_t), intent(inout) :: this
    integer, intent(in) :: idx

    integer :: i
    integer :: npts, nx, ny, nz
    character(len=100) :: filename
    integer :: ios, funit

    nx = this%nx + 1
    ny = this%ny + 1
    nz = this%nz + 1
    npts = nx * ny * nz

    ! Synchronize the device memory if using a GPU backend is enabled
    if (NEKO_BCKND_DEVICE .eq. 1) then
       call device_memcpy(this%x_coord%x, this%x_coord%x_d, npts, &
            DEVICE_TO_HOST, sync = .false.)
       call device_memcpy(this%y_coord%x, this%y_coord%x_d, npts, &
            DEVICE_TO_HOST, sync = .false.)
       call device_memcpy(this%z_coord%x, this%z_coord%x_d, npts, &
            DEVICE_TO_HOST, sync = .false.)
       call device_memcpy(this%values%x, this%values%x_d, npts, &
            DEVICE_TO_HOST, sync = .true.)
    end if

    write(filename, '(A,I4.4,A)') 'design_', idx, '.vtk'

    open(newunit=funit, file=filename, status='replace', action='write', iostat=ios)
    if (ios .ne. 0) then
       print *, 'Error opening file ', filename
       stop
    end if

    ! Header
    write(funit, '(A)') '# vtk DataFile Version 3.0'
    write(funit, '(A)') 'Simple Scalar Field'
    write(funit, '(A)') 'ASCII'
    write(funit, '(A)') 'DATASET STRUCTURED_GRID'
    write(funit, '(A,3(1X,I0))') 'DIMENSIONS', nx, ny, nz

    ! Points
    write(funit, '(A,1X,I0,1X,A)') 'POINTS', npts, 'float'
    do i = 1, npts
       write(funit, '(3(F20.12,1X))') this%x_coord%x(i), this%y_coord%x(i), &
            this%z_coord%x(i)
    end do

    ! Scalars
    write(funit, '(A,1X,I0)') 'POINT_DATA', npts
    write(funit, '(A)') 'SCALARS density float 1'
    write(funit, '(A)') 'LOOKUP_TABLE default'
    do i = 1, npts
       write(funit, '(F20.12)') this%values%x(i)
    end do

    close(funit)
  end subroutine design_simple_write
```

sources/designs/design_simplefield.f90
```diff
@@ -0,0 +1,225 @@
...
+  use fld_file_output, only: fld_file_output_t
+
+  implicit none
+  private
+
+  !> A topology optimization design variable
+  type, extends(design_t), public :: simplefield_design_t
+     private
+
+     type(field_t), private :: designfield
```
timfelle: As written two lines above. Everything is private unless marked public
```suggestion
     type(field_t) :: designfield
```

sources/designs/design_simplefield.f90
```diff
@@ -0,0 +1,225 @@
...
+     type(vector_t) :: x_coord
+     type(vector_t) :: y_coord
+     type(vector_t) :: z_coord
+
+     ! needed to write the design
+     type(fld_file_output_t), private :: output
```
timfelle:
```suggestion
     type(fld_file_output_t) :: output
```

examples/1d_beam/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Simple setup to compile the current example.
+set(DRIVER_TYPE "custom1")
```
timfelle:
```suggestion
set(DRIVER_TYPE "custom")
```

timfelle (changes requested):
Another batch of comments.

There seem to be something fishy with the 1d_beam plot function.

sources/designs/design_simple.f90
```diff
@@ -253,6 +274,59 @@ subroutine design_simple_write(this, idx)
...
+       call device_memcpy(this%values%x, this%values%x_d, npts, &
+            DEVICE_TO_HOST, sync = .true.)
+    end if
+
+    write(filename, '(A,I4.4,A)') 'design_', idx, '.vtk'
+
+    open(newunit=funit, file=filename, status='replace', action='write', iostat=ios)
+    if (ios .ne. 0) then
+       print *, 'Error opening file ', filename
+       stop
```
timfelle: The "stop command is not MPI safe. Please use neko error instead.
Also, print does not write to stderr so you have no guarantee the user will see
your message.
```suggestion
       call neko_error('Error opening file ' // filename)
```

sources/optimizer/mma.f90
```diff
@@ -337,14 +338,25 @@ subroutine mma_init_from_components(this, x, n, m, a0, a, c, d, xmin, xmax, &
...
+       call device_memcpy(this%a%x, this%a%x_d, m, HOST_TO_DEVICE, &
+            sync = .false.)
+       call device_memcpy(this%c%x, this%c%x_d, m, HOST_TO_DEVICE, &
+            sync = .false.)
+       call device_memcpy(this%d%x, this%d%x_d, m, HOST_TO_DEVICE, &
+            sync = .false.)
+       call device_memcpy(this%xmax%x, this%xmax%x_d, n, HOST_TO_DEVICE, &
+            sync = .false.)
+       call device_memcpy(this%xmin%x, this%xmin%x_d, n, HOST_TO_DEVICE, &
+            sync = .false.)
```
timfelle: Last memcpy of a block should always be sync. Otherwise you do not have a
guarantee that data has arrived on the GPU before the next command is done.
```suggestion
            sync = .true.)
```

examples/1d_beam/1d_beam.case
```diff
@@ -69,56 +58,30 @@
...
+    },
+    "optimization": {
+        "solver": {
+            "type": "mma",
+            "max_iterations": 6,
+            "tolerance": 1.0e-4,
+            "enable_output": true,
+            "mma": {
+                "max_iter": 100,
+                "backend": "cpu",
```
timfelle: Why not let the default be whatever we compiled with?
```suggestion
```

sources/optimizer/mma_optimizer.f90
```diff
@@ -227,17 +267,24 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
...
-       ! Stamp the i^th iteration
-       call mma_logger_assemble_data(log_data, iter, objective_value, &
-            all_objectives, constraint_value, this%mma%get_residumax(), &
-            this%mma%get_residunorm(), scaling_factor, &
-            problem%get_n_objectives(), problem%get_n_constraints())
-       call this%logger%write(log_data)
-
-       if (present(simulation)) call simulation%write(iter)
-       call design%write(iter)
+       call profiler_start_region("Optimizer logging")
```
timfelle: Please put this inside the if statement. Otherwise it will shop up with 0 time in the output.

timfelle (changes requested):
Segfaults for CUDA backend

timfelle (dismissed):
> Segfaults for CUDA backend 
Disregard. Was an issue in Neko.

sources/problems/constraints/volume_constraint.f90
timfelle:
```suggestion
```

sources/problems/problem.f90
```diff
@@ -632,4 +632,5 @@ function problem_get_log_header(this) result(buff)
     end do
 
   end function problem_get_log_header
+
```
timfelle:
```suggestion
```

timfelle:
@gloopydoop Please accept if you agree this is ready. I have set the automerge.

gloopydoop (approved):
yehaw :cowboy_hat_face: 
Great work boys :+1:

PR #287, opened 2025-09-24
Missing q in RAMP

gloopydoop:
Wow... we missed this one as well since RAMP always had q=1.
Added the fix and added RAMP up and down to the unit tests.
Also only check the tolerance for the smallest perturbation.


gloopydoop:
There's also a hug bug in the mapping cascade which has been fixed here

tests/unit/sensitivity/sensitivity.f90
```diff
@@ -213,7 +213,7 @@ subroutine compute_problem_sensitivity_i(problem, sim, des, target_sensitivities
        log_data%x(4) = fd_error
        call logger%write(log_data)
 
-       if (abs(fd_error) .gt. tolerance) then
+       if ((abs(fd_error) .gt. tolerance) .and. ip .eq. n_perturbations) then
```
timfelle, line 216: Why do you only test the final perturbation?

gloopydoop: RAMP is non-linear, so you don't get machines precision you get first order convergence.
image

timfelle: We do not test machine precision. That is why we test against a predefined tolerance to make sure we do not break the solver. That was why the tolerance was scaled by the perturbation in the driver I wrote.

If you want to test the linear convergence, then estimate the slope and test that the slope behaves as expected, then throw an error if that is violated.

If you only test the error on the last iteration, we do not know if something breaks later.

gloopydoop: based on those plots, and the perturbations chosen, that's 10^-1, which is pretty lax, I feel like anything could pass that test.

Testing the slope wouldn't work for the volume and filtered volume.

scaling by the perturbation kills you on the roundoff side (that would be particularly visible in the volume filtered plot, which would fail)

timfelle: What i did was assign a tolerance for each perturbation, which in the first case was defined by `base_tol * (perturbation_i / min(perturbation)` Which means tighter tolerances for smaller perturbations, and it was then tuned for the specific constraint function.

If you cannot make a universal rule, then you cannot make a universal test.

gloopydoop: image

This is the plot most people think of with a FD check. If we simply want these unit tests to catch breaking changes, a single point should suffice.

gloopydoop:
This was closed by #291 and #292

PR #296, opened 2025-10-07
Add profiler regions for checkpoints and MMA optimizer

timfelle:
Introduce profiler regions to enhance performance tracking during checkpoint saving, restoring, and MMA optimization iterations. This change aims to provide better insights into the execution time of critical operations.


gloopydoop (approved):
:+1:

PR #298, opened 2025-10-07
Unsteady adjoints

gloopydoop:
This PR is a trimmed version of the branch `unsteady/unsteady_adjoint` which contains much of #247 etc, but I think it stands alone.

The main contribution are: 
- tools to not just compute an objective value or sensitivity on a single field as is done in a steady simulation, eg
$F = \int_\Omega f(\mathbf{u}) d\Omega $
but now for time integrals, eg,
$F = \int^T_0 \int_\Omega f(\mathbf{u}(t)) d\Omega dt$

- a driver for the unsteady adjoint. Note, originally this was intended to be a part of simulation.f90, but trying to inject the problem back into the simulation led to cyclic dependencies. The unsteady adjoint loop really combines all of the three big players: simulation, problem and design.

PS, obviously `plot_sim_time_segments.py` was heavily vibe coded so we may want it removed from the repo, but it shows the checkpointing and adjoint computation seem reasonable:
Screenshot from 2025-10-07 14-33-57


gloopydoop:
superseded by #316

PR #299, opened 2025-10-08
Add GPU support for sensitivity

gloopydoop:
This PR ensure's the CPU and GPU versions of #247 are consistent.

here is a comparison between CPU and GPU (note the number of elements were slightly higher for the GPU version)

The key takeaways, the minimum dissipation is still slightly broken and I've tried EVERYTHING. But the scalar (which in the end is the important one) and lube term look excellent.

image

I've also found an indexing miss match in the masks which I remember being very conscious of at the time, but it appears this inconsistency was fixed on the neko side?


tests/regression/sensitivity/user.f90
```diff
@@ -72,6 +73,12 @@ subroutine user_bc(fields, bc, time)
           w%x(idx, 1, 1, 1) = 0.0_rp
        end do
 
+       if (NEKO_BCKND_DEVICE .eq. 1) then
+          call device_memcpy(u%x, u%x_d, u%size(), HOST_TO_DEVICE, sync=.false.)
+          call device_memcpy(v%x, v%x_d, v%size(), HOST_TO_DEVICE, sync=.false.)
+          call device_memcpy(w%x, w%x_d, w%size(), HOST_TO_DEVICE, sync=.false.)
```
timfelle: Last memcpy of a block should always be synced, otherwise you cannot be sure
that the memory arrived before it is being used next time
```suggestion
          call device_memcpy(w%x, w%x_d, w%size(), HOST_TO_DEVICE, sync=.true.)
```

tests/regression/sensitivity/user.f90
```diff
@@ -81,6 +88,9 @@ subroutine user_bc(fields, bc, time)
           ! Inflow scalar profile is a bump
           s%x(idx, 1, 1, 1) = 0.5_rp * (1.0_rp - cos(2.0_rp*pi*y))
        end do
+       if (NEKO_BCKND_DEVICE .eq. 1) then
+          call device_memcpy(s%x, s%x_d, s%size(), HOST_TO_DEVICE, sync=.false.)
```
timfelle:
```suggestion
          call device_memcpy(s%x, s%x_d, s%size(), HOST_TO_DEVICE, sync=.true.)
```

tests/regression/sensitivity/user.f90
```diff
@@ -101,6 +111,9 @@ subroutine scalar_ic(scheme_name, fields)
        s%x(i, 1, 1, 1) = 0.5_rp * (1.0_rp - cos(2.0_rp*pi*y))
     end do
 
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(s%x, s%x_d, s%size(), HOST_TO_DEVICE, sync=.false.)
```
timfelle:
```suggestion
       call device_memcpy(s%x, s%x_d, s%size(), HOST_TO_DEVICE, sync=.true.)
```

timfelle (commented):
The sensitivity "test" still does not test anything.
Please atleast compare against the reference CSV or something to make sure the test return a non-zero value if the finite differences are violated.

timfelle (approved):

PR #301, opened 2025-10-08
Reset adjoint completely

gloopydoop:
This PR allows the adjoint to be reset, including lags etc.

Depends on #300

### DIFF
--- a/sources/neko_ext/neko_ext.f90
+++ b/sources/neko_ext/neko_ext.f90
@@ -8,7 +8,7 @@
 !! optimization code.
 module neko_ext
   use case, only: case_t
-  use json_utils, only: json_get, json_get_or_default, json_extract_item
+  use json_utils, only: json_get, json_get_or_default
   use num_types, only: rp
   use simcomp_executor, only: neko_simcomps
   use flow_ic, only: set_flow_ic
@@ -26,6 +26,7 @@ module neko_ext
   use json_module, only : json_file
   use scalars, only: scalars_t
   use adjoint_scalars, only: adjoint_scalars_t
+  use field_math, only: field_rzero
 
   implicit none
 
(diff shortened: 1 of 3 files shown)


sources/neko_ext/neko_ext.f90
```diff
@@ -189,6 +186,140 @@ subroutine reset(neko_case)
...
+    ! Reset the timing parameters
+    ! ------------------------------------------------------------------------ !
+
+    t = 0.0_rp
+    adjoint_case%time%t = t
+    adjoint_case%time%tstep = 0
+
+    ! Setup lagged time step parameters
+    adjoint_case%time%tlag = t
+    adjoint_case%time%dtlag = adjoint_case%time%dt
```
timfelle: Please just reset the time state instead of manually assigning it to 0.

```suggestion
    call adjoint_case%time%reset()
    t = adjoint_case%time%start_time
```

sources/neko_ext/neko_ext.f90
```diff
@@ -184,6 +181,135 @@ subroutine reset(neko_case)
...
 
+!> @brief Reset the adjoint case data structure
+  !>
+  !> @details This subroutine resets the adjoint case data structure. It is
+  !> called at the beginning of each iteration. It is used to reset the time
+  !> step counter, the lagged time step parameters, the external BDF
+  !> coefficients, the adjoint fluid_adj and adjoint scalar fields.
+  !>
+  !> @param[inout] adjoint_case Adjoint case data structure.
+  !> @param[in] neko_case Primal case.
```
timfelle:
```suggestion
  !> @brief Reset the adjoint case data structure
  !!
  !! @details This subroutine resets the adjoint case data structure. It is
  !! called at the beginning of each iteration. It is used to reset the time
  !! step counter, the lagged time step parameters, the external BDF
  !! coefficients, the adjoint fluid_adj and adjoint scalar fields.
  !!
  !! @param[inout] adjoint_case Adjoint case data structure.
  !! @param[in] neko_case Primal case.
```

timfelle (approved):

PR #303, opened 2025-10-10
Re-add missing n_steps counter and fix region names

timfelle:
Restore the n_steps counter in the simulation run and correct the names of profiling regions for clarity.


gloopydoop (approved):
nice :sweat_smile:

PR #305, opened 2025-10-28
Introduce the mixer benchmark

timfelle:
Introduce the mixer benchmark case.


bench/mixer/templates/LUMI-G.sh
```diff
@@ -0,0 +1,71 @@
...
+
+export ROCR_VISIBLE_DEVICES=\$SLURM_LOCALID
+exec \$*
+EOF
+
+chmod +x ./select_gpu
+export CPU_BIND="map_cpu:49,57,17,25,1,9,33,41"
+export MPICH_GPU_SUPPORT_ENABLED=1
+
+lfs setstripe -S 2m -c 72 ./
```
gloopydoop, line 67: It's been a while since the LUMI course :sweat_smile:  isn't striping very case specific?

gloopydoop (approved):
:+1: 
I've clicked approved, only one comment about striping, but it's a bit of a moot point since it's on a template

bench/mixer/templates/LUMI-G.sh
timfelle: Yes it is. I did not study it in exact detail, but this is what was used to do our benchmark for the application. Which I think should be reflected here until we can explain something more precise.

PR #306, opened 2025-10-28
Introduce example of a static mixer

timfelle:


examples/static_mixer/petsc/case.template
```diff
@@ -0,0 +1,106 @@
+{
+    "version": 1.0,
+    "case": {
+        "mesh_file": "data_local/static_mixer/mixer_8x2x2.nmsh",
```
gloopydoop, line 4: Is there a reason this isn't created with a `prepare.sh`?

examples/static_mixer/petsc/mixer.f90
```diff
@@ -0,0 +1,143 @@
...
+       s => fields%items(1)%ptr
+       call scalar_z_split_ic(s, 0.5_rp, 0.0_rp, 1.0_rp)
+    end if
+
+    if (scheme_name .eq. 'fluid') then
+       u => fields%get("u")
+       v => fields%get("v")
+       w => fields%get("w")
+       call field_cfill(u, 1.0_rp)
+       call field_subcol3(u, u, brinkman_indicator)
```
gloopydoop, line 45: This is nice. An alternative is to use the brinkman indicator as a mask and zero out everything in the solid as an initial condition. We should try and migrate these kinds of nice touches somewhere into main source since it would be useful outside of this example

examples/static_mixer/petsc/mixer.f90
```diff
@@ -0,0 +1,143 @@
...
+       s => fields%get("temperature")
+
+       do i = 1, bc%msk(0)
+          z = bc%dof%z(bc%msk(i), 1, 1, 1)
+          if (z .gt. 0.5_rp) then
+             s%x(bc%msk(i), 1, 1, 1) = 1.0_rp
+          else
+             s%x(bc%msk(i), 1, 1, 1) = 0.0_rp
+          end if
+       end do
```
gloopydoop, line 110: perhaps a smooth transitional region? You already have the smooth step. In principal this could lead to Gibbs oscillations in the temperature

gloopydoop (approved):
Only a few comments, looks good :+1:

examples/static_mixer/petsc/case.template
timfelle: It is done in the create examples to make sure it is only done once.

examples/static_mixer/petsc/mixer.f90
timfelle: Added step from 0.45 to 0.55

PR #307, opened 2025-10-28
Refactor MMA optimizer profiling regions and checkpoint file handling

timfelle:
Remove duplicate includes, update profiling regions in the MMA optimizer for better performance tracking, and enhance the checkpoint class to delete files more effectively upon freeing.

### DIFF
--- a/sources/checkpoint/checkpoint.f90
+++ b/sources/checkpoint/checkpoint.f90
@@ -246,10 +246,11 @@ contains
        do i = this%n_timesteps, 1, -1
           call this%chkp_output%set_counter(i)
           file_name = this%chkp_output%file_%get_fname()
-          inquire(file=trim(file_name), exist=exists)
+          inquire(file = trim(file_name), exist = exists)
           if (exists) then
-             open(newunit=unit, file=trim(file_name), iostat=stat, status='old')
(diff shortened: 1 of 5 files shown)


sources/checkpoint/checkpoint.f90
```diff
@@ -239,10 +242,17 @@ subroutine checkpoint_free(this)
...
+       do i = this%n_timesteps, 1, -1
+          call this%chkp_output%set_counter(i)
+          file_name = this%chkp_output%file_%get_fname()
+          inquire(file = trim(file_name), exist = exists)
+          if (exists) then
+             open(newunit = unit, file = trim(file_name), iostat = stat, &
+                  status='old')
+             if (stat .eq. 0) close(unit, status = 'delete')
+          end if
+       end do
```
gloopydoop, line 255: this is cool!
If I'm understanding correctly, this cleans out all the flds upon completion. Important, nice :+1:

gloopydoop (approved):
looks good :+1:

sources/checkpoint/checkpoint.f90
timfelle: yes exactly

PR #311, opened 2025-11-10
Add MMA optimizer tests and update scripts

timfelle:
Introduce a new regression test suite for the MMA optimizer, including necessary scripts and configuration files. Update existing scripts and references to enhance functionality and maintainability.


AbbasMoosaV (commented):
Everything looks great.
Maybe reduce the tolerance in check.py.
Currently, I get "Discrepancies found compared to reference data" because 3 values are slightly larger than E-10 (1.17471380e-10).

AbbasMoosaV (approved):

PR #314, opened 2025-11-11
target dissipation objective

gloopydoop:
Introduces a target dissipation objective that can ideally be used like a pressure drop constraint.
So far tested and gives results similar to the minimum dissipation (which gets fixed with the new adjoint calculation method)
image

Depends on #368


sources/problem/objectives/target_dissipation_objective.f90
```diff
@@ -0,0 +1,290 @@
...
+    type(simulation_t), target, intent(inout) :: simulation
+
+    character(len=:), allocatable :: name
+    character(len=:), allocatable :: mask_name
+    real(kind=rp) :: weight, target_fraction
+
+    call json_get_or_default(json, "weight", weight, 1.0_rp)
+    call json_get_or_default(json, "mask_name", mask_name, "")
+    call json_get_or_default(json, "name", name, "Target Dissipation")
+    call json_get(json, "target", target_fraction)
```
AbbasMoosaV, line 176: I think we should provide a default value for "target" here. (e.g. 0.5_rp)

gloopydoop: open to the idea, but I certainly wouldn't suggest 0.5 or anything <1.

I originally left it with no default since it's very case dependent. Remember it's expressed as a percentage of the first iteration, so if you start with an empty channel it's pretty hard to design something with LESS dissipation. Maybe 2 would make sense?

timfelle (commented):
@AbbasMoosaV @gloopydoop what is the status here?

AbbasMoosaV:
> @AbbasMoosaV @gloopydoop what is the status here?

Well I have tested this objective with the steady mixer and I think it looks good.
So, @gloopydoop if you are happy with this, we can update this with develop and merge it.

sources/source_terms/adjoint_target_dissipation_source_term.f90
```diff
@@ -0,0 +1,322 @@
+! Copyright (c) 2023, The Neko Authors
```
timfelle: Please check the dates for copyright

sources/problem/objectives/target_dissipation_objective.f90
```diff
@@ -0,0 +1,361 @@
...
+    this%current_dissipation = this%value
+
+    ! Now compute the objective
+    this%value = 0.5_rp * (this%current_dissipation / &
+         (this%initial_dissipation * this%target_fraction) - 1.0_rp) ** 2
+
+  end subroutine target_dissipation_finalize_value
+
+  !> update_value the sensitivity of the objective function with respect to
+  !! \f\f$\chi\f\f$
```
timfelle:
```suggestion
  !! \f$\chi\f$
```

sources/problem/objectives/target_dissipation_objective.f90
```diff
@@ -0,0 +1,361 @@
...
+    class(target_dissipation_objective_t), intent(in) :: this
+    integer :: n
+
+    n = 5
+  end function target_dissipation_get_log_size
+
+  !> Header labels for log entries
+  subroutine target_dissipation_get_log_headers(this, headers)
+    class(target_dissipation_objective_t), intent(in) :: this
+    character(len=*), intent(out) :: headers(:)
```
timfelle, line 517:
```suggestion
    character(len=*), intent(inout) :: headers(:)
```
You allocated this outside and fill it here. Then you cannot have this as an out. Otherwise it would be deallocated.

sources/problem/objectives/target_dissipation_objective.f90
```diff
@@ -0,0 +1,361 @@
...
+    values(1) = this%value
+    if (size(values) .lt. 2) return
+    values(2) = this%weight
+    if (size(values) .lt. 3) return
+    values(3) = this%current_dissipation
+    if (size(values) .lt. 4) return
+    values(4) = this%initial_dissipation
+    if (size(values) .lt. 5) return
+    values(5) = this%current_dissipation / this%initial_dissipation
+
```
timfelle, line 557: You ahould probably do 'if (size...) headers(...' instead of the early returns. The compilers usually like that better than  the early returns.

sources/source_terms/adjoint_target_dissipation_source_term.f90
```diff
@@ -0,0 +1,322 @@
...
+  use scratch_registry, only: neko_scratch_registry
+  use mask_ops, only: mask_exterior_const
+  use point_zone, only: point_zone_t
+  use ax_product, only : ax_t, ax_helm_factory
+
+  implicit none
+  private
+
+  !> An adjoint source term for objectives of minimum dissipation
+  ! $\int \nabla v \cdot \nabla u $
```
timfelle:
```suggestion
  ! \f$\int \nabla v \cdot \nabla u \f$
```

sources/source_terms/adjoint_target_dissipation_source_term.f90
```diff
@@ -0,0 +1,322 @@
...
+    real(kind=rp), intent(in) :: volume
+    real(kind=rp), intent(in) :: target_fraction
+    real(kind=rp), target, intent(in) :: current_dissipation
+    real(kind=rp), target, intent(in) :: initial_dissipation
+    logical :: if_mask
+
+    ! I wish you didn't need a start time and end time...
+    ! but I'm just going to set a super big number...
+    start_time = 0.0_rp
+    end_time = 100000000.0_rp
```
timfelle: I am not sure you do. Please look at the source_term_t to make aure it is not optional

timfelle (changes requested):

gloopydoop:
This has been lingering forever, I'll keep the branch live but there's no need for a PR

PR #316, opened 2025-11-12
Unsteady adjoint (continuous in time)

gloopydoop:
This PR is a trimmed version of the branch `unsteady/unsteady_adjoint` which contains much of #247 etc, but I think it stands alone.

The main contribution are: 
- tools to not just compute an objective value or sensitivity on a single field as is done in a steady simulation, eg
$F = \int_\Omega f(\mathbf{u}) d\Omega $
but now for time integrals, eg,
$F = \int^T_0 \int_\Omega f(\mathbf{u}(t)) d\Omega dt$

- a driver for the unsteady adjoint. Note, originally this was intended to be a part of simulation.f90, but trying to inject the problem back into the simulation led to cyclic dependencies. The unsteady adjoint loop really combines all of the three big players: simulation, problem and design.

PS, obviously `plot_sim_time_segments.py` was heavily vibe coded so we may want it removed from the repo, but it shows the checkpointing and adjoint computation seem reasonable:


sources/problems/problem.f90
```diff
@@ -378,12 +429,24 @@ subroutine problem_compute_sensitivity(this, design, simulation)
     class(problem_t), intent(inout) :: this
     class(design_t), intent(inout) :: design
     class(simulation_t), optional, intent(inout) :: simulation
+    logical :: unsteady
 
     type(vector_t) :: objective_sensitivity
 
-    if (present(simulation)) call simulation%run_backward()
+    if (present(simulation)) then
+       call simulation%checkpoint%get_enabled(unsteady)
```
timfelle: Please move this to the simulation. A checkpoint cannot be unsteady.

timfelle (changes requested):

sources/optimizer/mma_optimizer.f90
```diff
@@ -201,7 +201,13 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
     call profiler_start_region("Optimizer iteration")
 
     call problem%compute(design, simulation)
+    if (present(simulation) .and. this%enable_output) then
+       call simulation%write_forward(0)
+    end if
     call problem%compute_sensitivity(design, simulation)
+    if (present(simulation) .and. this%enable_output) then
+       call simulation%write_adjoint(0)
+    end if
```
timfelle, line 219: Why is this split. and why did you move it out of the logging block?

sources/optimizer/mma_optimizer.f90
```diff
@@ -266,7 +268,13 @@ subroutine mma_optimizer_run(this, problem, design, simulation)
        call design%update_design(x)
 
        call problem%compute(design, simulation)
+       if (present(simulation) .and. this%enable_output) then
+          call simulation%write_forward(iter)
+       end if
        call problem%compute_sensitivity(design, simulation)
+       if (present(simulation) .and. this%enable_output) then
+          call simulation%write_adjoint(iter)
+       end if
```
timfelle, line 285: Again here.

sources/checkpoint/checkpoint.f90
```diff
@@ -98,6 +98,8 @@ module simulation_checkpoint
      procedure, public, pass(this) :: save => checkpoint_save
      !> Restore the forward simulation state
      procedure, public, pass(this) :: restore => checkpoint_restore
+     !> Public getter to check if the simulation is unsteady
+     procedure, public, pass(this) :: get_enabled => checkpoint_get_enabled
```
timfelle: This is not the way to do this. A simulation is not unsteady because it have a
checkpoint structure. An unsteady simulation could also not have one, hence our
entire POD idea.

sources/problems/problem.f90
```diff
@@ -394,6 +457,100 @@ subroutine problem_compute_sensitivity(this, design, simulation)
...
+    !  - n    time steps to compute
+    !  - n+1  fields to consider
+    !  - n-1  non-zero contributions to u * u_adj
+    !
+    !              non-zero
+    !             |--------|
+    !  primal  o--x--x--x--x--x
+    !          x--x--x--x--x--o  adjoint
+    !          ^              ^
+    !         u=0          u_adj=0
```
timfelle, line 531: what are you on about. we do have a checkpoint at i = 0, i = 1 and i = 2. Since
any simulation starting from either 0 or 1 does not match the solution so we
need to manually load them, which we do. You initialize a checkpoint structure
with a      integer :: first_valid_timestep = 2. All timesteps before that are
directly loaded from disc and will not be resimulated.

sources/problems/problem.f90
```diff
@@ -394,6 +457,100 @@ subroutine problem_compute_sensitivity(this, design, simulation)
...
+       ! accumulate objective value
+       call this%accumulate_objectives(design, simulation%adjoint_case%time%dt)
+       ! save a checkpoint
+       call simulation%checkpoint%save(simulation%neko_case)
+    end do
+    call profiler_end_region("Forward simulation")
+
+    call simulation_finalize(simulation%neko_case)
+
+  end subroutine problem_run_forward_unsteady
```
timfelle, line 489: This is a simulation loop. Not a problem loop. It does not belong here.

sources/checkpoint/checkpoint.f90
```diff
@@ -322,6 +324,15 @@ subroutine checkpoint_restore(this, neko_case, tstep)
...
 
+  !> Restore the forward simulation state
+  subroutine checkpoint_get_enabled(this, enabled)
+    class(simulation_checkpoint_t), intent(inout) :: this
+    logical, intent(out) :: enabled
+
+    enabled = this%enabled
+
+  end subroutine checkpoint_get_enabled
+
```
timfelle: Delete

sources/simulation/simulation.f90
```diff
@@ -288,4 +292,22 @@ subroutine simulation_write(this, idx)
```
timfelle, line 317: Please call the internal write forward and write backward instead of duplicating code

sources/optimizer/mma_optimizer.f90
gloopydoop: The split is because if you run forward, then backward, then write the fields, you're writing the initial forward field which is just the IC and will be the same across all optimization iterations. The final field forward field is the one that's interesting.

sources/problems/problem.f90
gloopydoop: I'm saying:
$`\int_0^T v \cdot \frac{\partial \delta u}{\partial t} dt = |v \cdot \delta u|_{t=T}  - |v \cdot \delta u|_{t=0} - \int_0^T \frac{\partial v} {\partial t} \cdot \delta u dt `$

So those two "boundary" (in time) terms cancel with $`u|_{t=0} = 0`$ and $v|_{t=T} = 0$. Which is standard. I'm pretty sure there are circumstances where this isn't the case and special care needs to be taken here (optimal initial conditions?). But really I don't know, this comment was mostly there to say this only works if we start from $`u|_{t=0} = 0`$ and $`v|_{t=T} = 0`$

sources/problems/problem.f90
gloopydoop: man, this one was tough one, I agree completely. I would have loved this in the simulation and that's what I tried first, but we need the `problem_t` to compute the objective and accumulate the sensitivity, so trying to inject that into `simulation_t` led to cyclic dependencies. If you can figure out a good way to resolve that, I agree getting this into the simulation would be good.

sources/problems/problem.f90
gloopydoop: ba02851d59c00a03115a6a7e3f76defed5fe9460

sources/checkpoint/checkpoint.f90
gloopydoop: ba02851d59c00a03115a6a7e3f76defed5fe9460

sources/checkpoint/checkpoint.f90
gloopydoop: ba02851d59c00a03115a6a7e3f76defed5fe9460

timfelle (changes requested):
The unsteady_ryugby_ball example included here does not run.
```
unsteady_rugby_ball_____________________________________________________________
 *** ERROR: Requested timestep 2000 is out of range [1, 1000] ***
```

gloopydoop:
> The unsteady_ryugby_ball example included here does not run.
> 
> ```
> unsteady_rugby_ball_____________________________________________________________
>  *** ERROR: Requested timestep 2000 is out of range [1, 1000] ***
> ```

ed42a4d0a89381671ef41cf27f73b44683f5d40f

sources/problems/base_functional.f90
```diff
@@ -164,4 +174,47 @@ subroutine functional_get_sensitivity(this, sensitivity)
...
+    ! should suffice
+    this%value = temp1 + temp2 * dt
+  end subroutine functional_accumulate_value
+
+  !> Accumulate the value of the function
+  subroutine functional_accumulate_sensitivity(this, design, dt)
+    class(base_functional_t), intent(inout) :: this
+    class(design_t), intent(in) :: design
+    real(kind=rp), intent(in) :: dt
+    type(vector_t) :: temp
```
timfelle, line 212: temp is never freed

timfelle (changes requested):
The design in unsteady_rugby_ball is never updated. Design at iteration 50 is identical to the one at iteration 0

sources/problems/base_functional.f90
gloopydoop: 093185e4e3dc285d16621b134ce30ec41b460d08

But I guess this will be handled more elegantly after neko  #2165

gloopydoop:
> The design in unsteady_rugby_ball is never updated. Design at iteration 50 is identical to the one at iteration 0

nice catch :+1: 

118bf662807903e5f77200dd958387139c06c1d9

It's not the best example due to symmetry breaking etc...

timfelle (approved):

examples/unsteady_rugby_ball/user.f90
```diff
@@ -0,0 +1,60 @@
+! User module for the user defined simulation component
+module user
+  use user_intf, only: user_t
+  use num_types, only : rp
+  use field, only : field_t
+  use field_list, only : field_list_t
+  use field_dirichlet, only : field_dirichlet_t
+  use time_state, only : time_state_t
+  use field_registry, only : neko_field_registry
```
timfelle:
```suggestion
```

PR #317, opened 2025-11-19
Align with new scratch interface of Neko

timfelle:
Update the code to use the new `field_scratch_registry` interface instead of the deprecated `scratch_registry`. This change ensures compatibility with the latest Neko framework updates.

Depends on #315


gloopydoop (approved):
looks good to me! And good work on the  neko side enabling this :+1:

PR #342, opened 2026-01-13
Enhance MPI functionality and fix bugs

timfelle:
Improve MPI integration by adding global reductions and checkpointing capabilities. Resolve various bugs in the reader and writer, and ensure regression tests pass successfully. Clean up code and simplify certain operations for better performance.

### DIFF
--- a/sources/mma/bcknd/cpu/mma_cpu.f90
+++ b/sources/mma/bcknd/cpu/mma_cpu.f90
@@ -721,6 +721,8 @@ contains
 
              residual_small = [rey, rez, relambda, remu, rezeta, res]
              new_residual = sqrt(norm2(residual_small)**2 + re_sq_norm)
+             call MPI_Allreduce(MPI_IN_PLACE, new_residual, 1, &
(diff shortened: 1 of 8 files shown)


gloopydoop (approved):
This looks good! I didn't realize there was so much required for the IO, nice job. 
Small question: do we have room in hdf5 for some kind of metadata as well within the checkpoints? Right now I'm just thinking something as simple as optimization iteration number (I guess we have the name of the file...) but in the future, with things like continuation strategies etc, it might be a nice option. :+1:

timfelle (commented):
> This looks good! I didn't realize there was so much required for the IO, nice job. 
> Small question: do we have room in hdf5 for some kind of metadata as well within the checkpoints? Right now I'm just thinking something as simple as optimization iteration number (I guess we have the name of the file...) but in the future, with things like continuation strategies etc, it might be a nice option. :+1:  

Well the sky is the limit. But yes, the optimizer will insert the iteration number so that when restarted we actually restart. But for the MMA object, the iteration number is not relevant for a restart so thats why it is not there. We will not rely on the file name to track the iteration number in general.

PR #366, opened 2026-02-16
Refactor MMA scratch registry usage; update device and core MMA modules

AbbasMoosaV:
Create a local scratch registry for mma_t.

### DIFF
--- a/sources/mma/mma.f90
+++ b/sources/mma/mma.f90
@@ -95,7 +95,7 @@ module mma
      type(vector_t) :: xold1, xold2, low, upp, alpha, beta, a, c, d, xmax, xmin
      logical :: is_initialized = .false.
      logical :: is_updated = .false.
-     type(scratch_registry_t) :: mma_scratch_registry
+     type(scratch_registry_t) :: scratch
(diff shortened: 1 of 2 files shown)


sources/mma/mma.f90
```diff
@@ -94,6 +95,7 @@ module mma
      type(vector_t) :: xold1, xold2, low, upp, alpha, beta, a, c, d, xmax, xmin
      logical :: is_initialized = .false.
      logical :: is_updated = .false.
+     type(scratch_registry_t) :: mma_scratch_registry
```
timfelle: just call it `scratch_registry` we know it is for mma since it lives inside the object.
```suggestion
     type(scratch_registry_t) :: scratch_registry
```

timfelle (changes requested):
You forgot to update the CPU codes to use the registry.

sources/mma/mma.f90
AbbasMoosaV: But we have a module in Neko called  `scratch_registry`.
In my experience, this will give errors on Cray machines.
You cannot have a module and object with the same name for some reason.

timfelle (approved):

PR #368, opened 2026-02-18
Improve objective/constraint logging

gloopydoop:
Currently each objective and constraint can only put their value into optimization_data.txt.

I thought it would be nice they had to power to put extra values in to make post processing easier. 
Right now it's only really the 
- volume constraint where this is useful :sweat_smile: 
- dummy_constraint (to show nothing)
- augmented_lagrangian (to show nothing)

But it also also now displays the value and the scaled value of objectives, making plotting multi objectives easier

But #314 will benefit from this


sources/optimizer/optimizer.f90
```diff
@@ -277,6 +289,12 @@ subroutine optimizer_free_base(this)
 
     this%start_time = 0.0_rp
     this%current_iteration = 0
+    if (this%log_initialized) then
+       call this%log_data%free()
+    end if
```
timfelle: It is always safe to call free.
```suggestion
    call this%log_data%free()
```

sources/optimizer/optimizer.f90
```diff
@@ -67,6 +69,12 @@ module optimizer
      real(kind=rp), private :: start_time = 0.0_rp
      real(kind=rp), private :: average_time = 0.0_rp
      real(kind=rp), private :: step_count = 0.0_rp
+     !> Logging state
```
timfelle: This comment only attaches to `log_initialized` Please only use `!>` when you
actually wish to attach the comment to a specific object.

sources/problem/problem.f90
```diff
@@ -867,37 +918,83 @@ pure function problem_get_num_constraints(this) result(n)
...
+
+  !> Return the base log size (excluding iter and optimizer extras).
+  !! @param[in] this The problem object.
+  !! @param[in] include_constraints Include constraints in the size count.
+  !! @return n Number of log entries excluding iteration and optimizer extras.
+  function problem_get_log_size(this, include_constraints) result(n)
+    class(problem_t), intent(in) :: this
+    logical, intent(in), optional :: include_constraints
+    integer :: n
+    integer :: i
```
timfelle:
```suggestion
    integer :: n, i
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+    class(optimizer_t), intent(inout) :: this
+    integer, intent(in) :: iter
+    class(problem_t), intent(in) :: problem
+    real(kind=rp), intent(in), optional :: extra_values(:)
+    integer :: base_size, offset, j
+
+    if (.not. this%log_initialized) return
+
+    base_size = problem%get_log_size(include_constraints = &
+         this%log_include_constraints)
```
timfelle: There is not reason to do the `=` thing
```suggestion
    base_size = problem%get_log_size(this%log_include_constraints)
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+    this%log_data%x = 0.0_rp
+    this%log_data%x(1) = real(iter, kind=rp)
+
+    call problem%get_log_values( &
+         this%log_data%x(2:1 + base_size), &
+         include_constraints = this%log_include_constraints)
```
timfelle: Same here, you do not help the reader.
We often use the equality when it is ambiguous which parameter is meant, or when
you only want to specify some of the optional parameters.
```suggestion
         this%log_include_constraints)
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+                  extra_values
+          else
+             do j = 0, this%log_extra_size - 1
+                this%log_data%x(offset + j) = 0.0_rp
+             end do
+          end if
+       else
+          do j = 0, this%log_extra_size - 1
+             this%log_data%x(offset + j) = 0.0_rp
+          end do
```
timfelle: This should not be needed. You have already pre-filled the entire array with
0's.
```suggestion
```

sources/optimizer/optimizer.f90
```diff
@@ -67,6 +69,12 @@ module optimizer
      real(kind=rp), private :: start_time = 0.0_rp
      real(kind=rp), private :: average_time = 0.0_rp
      real(kind=rp), private :: step_count = 0.0_rp
+     !> Logging state
+     logical, private :: log_initialized = .false.
+     logical, private :: log_include_constraints = .true.
+     integer, private :: log_extra_size = 0
+     type(csv_file_t), private :: log_file
+     type(vector_t), private :: log_data
```
timfelle, line 89: Why do you need to place the log data on the device?

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+         this%log_include_constraints)
+    this%log_data%x = 0.0_rp
+    this%log_data%x(1) = real(iter, kind=rp)
+
+    call problem%get_log_values( &
+         this%log_data%x(2:1 + base_size), &
+         include_constraints = this%log_include_constraints)
+
+    offset = 2 + base_size
+    if (this%log_extra_size .gt. 0) then
```
timfelle: Why do you need this extra if statement? if the `extra_values` are not present,
then the size should be 0, so why do you need both?

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+       if (present(extra_values)) then
+          if (size(extra_values) .eq. this%log_extra_size) then
+             this%log_data%x(offset:offset + this%log_extra_size - 1) = &
+                  extra_values
+          else
```
timfelle: This can be simplified quite a bit.
```suggestion
       if (present(extra_values)) then
          this%log_data%x(offset:offset + size(extra_values) - 1) = extra_values
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+    else
+       this%log_include_constraints = .true.
+    end if
+
+    base_size = problem%get_log_size(include_constraints = &
+         this%log_include_constraints)
+
+    extra_n = 0
+    if (present(extra_size)) extra_n = extra_size
+    this%log_extra_size = extra_n
```
timfelle: Way over complicated.
```suggestion
    if (present(extra_size)) this%log_extra_size = extra_size
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+
+    character(len=4096) :: header
+    integer :: total_size, extra_n, base_size
+    character(len=256) :: log_name
+
+    if (present(include_constraints)) then
+       this%log_include_constraints = include_constraints
+    else
+       this%log_include_constraints = .true.
+    end if
```
timfelle, line 555: Add the default value to the class variable
```suggestion
    if (present(include_constraints)) then
       this%log_include_constraints = include_constraints
    end if
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+
+    base_size = problem%get_log_size(include_constraints = &
+         this%log_include_constraints)
```
timfelle: The variable input parameter name is redundant, as you have a verbose variable name.
```suggestion
    base_size = problem%get_log_size(this%log_include_constraints)
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+  !! @param[in] include_constraints Include constraints in the log.
+  !! @param[in] filename Output filename for the log.
+  subroutine optimizer_init_log(this, problem, extra_headers, extra_size, &
+       include_constraints, filename)
+    class(optimizer_t), intent(inout) :: this
+    class(problem_t), intent(in) :: problem
+    character(len=*), intent(in), optional :: extra_headers
+    integer, intent(in), optional :: extra_size
+    logical, intent(in), optional :: include_constraints
+    character(len=*), intent(in), optional :: filename
```
timfelle, line 547: Why would you allow the user to add values without specifying the headers for
them?

```suggestion
  !> Initialize optimization log.
  !! @param[inout] this The optimizer object.
  !! @param[in] problem The problem object.
  !! @param[in] extra_headers Comma-separated extra header labels.
  !! @param[in] include_constraints Include constraints in the log.
  !! @param[in] filename Output filename for the log.
  subroutine optimizer_init_log(this, problem, extra_headers, &
       include_constraints, filename)
    class(optimizer_t), intent(inout) :: this
    class(problem_t), intent(in) :: problem
    character(len=*), intent(in), optional :: extra_headers
    logical, intent(in), optional :: include_constraints
    character(len=*), intent(in), optional :: filename
```

sources/optimizer/optimizer.f90
```diff
@@ -434,6 +452,109 @@ function optimizer_out_of_time(this, step_time) result(out_of_time)
...
+
+    call this%log_file%init(trim(log_name))
+
+    header = 'iter, ' // trim(problem%get_log_header( &
+         include_constraints = this%log_include_constraints))
+    if (present(extra_headers)) then
+       if (trim(extra_headers) .ne. '' .and. extra_n .gt. 0) then
+          header = trim(header) // ', ' // trim(extra_headers)
+       end if
+    end if
```
timfelle, line 582:
```suggestion
    if (present(extra_headers)) then
       if (trim(extra_headers) .eq. '') then
          call neko_error("Some headers are empty")
       end if
       header = trim(header) // ', ' // trim(extra_headers)
    end if
```

sources/optimizer/optimizer.f90
timfelle: Also why are the headers not an array? otherwise it is only a single header.

sources/problem/constraints/volume_constraint.f90
```diff
@@ -98,6 +98,13 @@ module volume_constraint
      !> Computes the source term and adds the result to `fields`.
      procedure, public, pass(this) :: update_sensitivity => &
           volume_constraint_update_sensitivity
+     !> Log sizing and values
```
timfelle: Why do you state that the `get_log_size` will return the logging values?

timfelle (approved):

PR #374, opened 2026-02-27
heaviside projection mapping

gloopydoop:
After playing around with the mixer I think this is going to be needed.
Implements a heaviside projection to the mapping cascade.
Still maps between [0,1] so should be used in conjunction with a linear mapping or so.


sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -0,0 +1,204 @@
...
+
+  !> Constructor from json.
+  subroutine heaviside_mapping_init_from_json(this, json, coef)
+    class(heaviside_mapping_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    type(coef_t), intent(inout) :: coef
+    real(kind=rp) :: beta, eta
+
+    call json_get_or_default(json, 'beta', beta, 8.0_rp)
+    call json_get_or_default(json, 'eta', eta, 0.5_rp)
```
timfelle: Where do you have these values from?

timfelle (changes requested):

sources/mapping_functions/PDE_filter_mapping.f90
timfelle, line 1: What do these changes have to do with the heaviside projection?

gloopydoop: @AbbasMoosaV 's commit.

sources/mapping_functions/heaviside_mapping.f90
gloopydoop: the eta one is natural, $\rho\in[0,1]$, so 0.5 makes sense.
Beta is arbitrary, and is generally used as part of a continuation strategy, by progressively increasing, so larger is sharper. I've seen people progressively double in their continuation strategy and  8 is my favourite power of 2, but I've read people do their continuation strategy from [1, 512] :grimacing: .

We can switch to a json_get for beta if you want, but I think a default of 0.5 for eta makes sense.

sources/mapping_functions/PDE_filter_mapping.f90
AbbasMoosaV: I will remove this from this PR.
We will add a new PR updating these indexing issues after we merge in the Heaviside stuff.

AbbasMoosaV (approved):

sources/mapping_functions/heaviside_mapping.f90
timfelle: \f$\eta = 0.5\f$ Makes sense as a default. But lets remove the default for \f$\beta\f$. As you mention there is no single "natural" value.

sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -0,0 +1,204 @@
...
+contains
+
+  !> Constructor from json.
+  subroutine heaviside_mapping_init_from_json(this, json, coef)
+    class(heaviside_mapping_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    type(coef_t), intent(inout) :: coef
+    real(kind=rp) :: beta, eta
+
+    call json_get_or_default(json, 'beta', beta, 8.0_rp)
```
timfelle:
```suggestion
    call json_get(json, 'beta', beta)
```

timfelle (approved):

sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -0,0 +1,204 @@
...
+  use field, only: field_t
+  use coefs, only: coef_t
+  use neko_config, only: NEKO_BCKND_DEVICE
+  use device_heaviside_mapping, only: &
+       device_heaviside_mapping_apply, &
+       device_heaviside_mapping_apply_backward
+  use heaviside_mapping_cpu, only: &
+       heaviside_mapping_apply_cpu, &
+       heaviside_mapping_apply_backward_cpu
+  use json_utils, only: json_get_or_default
```
timfelle:
```suggestion
  use json_utils, only: json_get, json_get_or_default
```

timfelle (approved):

PR #375, opened 2026-03-02
Verbose mapping cascade

gloopydoop:
This PR allows intermediate fields in the mapping cascade to be be visualized. It is not default behavior, because I view this more as a debugging option.

**NOTE**

This does change the output even in the default case. We used to have `design0.f****` holding 
u. unmapped
v. mapped
w. sensitivity
(which was not super clear even to begin with)

Now we have
`design0.f****`
p. unmapped
t. mapped

`sensitivity0.f****`
p. sensitivity

or, if you use the flags for verbose, then they will come in order, same thing for chain ruling backwards

### DIFF
--- a/sources/mapping_functions/mapping_handler.f90
+++ b/sources/mapping_functions/mapping_handler.f90
@@ -145,61 +145,61 @@ contains
     call this%sensitivity_output%free()
 
     if (allocated(this%sensitivity_stages)) then
-      do i = 1, size(this%sensitivity_stages)
-        call this%sensitivity_stages(i)%free()
-      end do
-      deallocate(this%sensitivity_stages)
+       do i = 1, size(this%sensitivity_stages)
+          call this%sensitivity_stages(i)%free()
+       end do
+       deallocate(this%sensitivity_stages)
     end if
 
     n_mappings = 0
     if (allocated(this%mapping_cascade)) n_mappings = size(this%mapping_cascade)
 
     if (this%verbose_design) then
-      ! all mapping inputs + mapped output
-      n_fields = n_mappings + 1
-      if (n_fields .le. 0) n_fields = 1
+       ! all mapping inputs + mapped output
+       n_fields = n_mappings + 1
+       if (n_fields .le. 0) n_fields = 1
     else
-      ! only first unmapped stage and mapped output
-      if (n_mappings .gt. 0) then
-        n_fields = 2
-      else
-        n_fields = 1
-      end if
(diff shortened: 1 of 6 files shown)


sources/mapping_functions/mapping_handler.f90
```diff
@@ -350,9 +486,71 @@ subroutine mapping_handler_add_mapping(this, mapping)
...
+  !! @param[in] verbose_sensitivity If true, output all backward stages.
+  !! @param[in] output_precision Output precision (sp or dp).
+  subroutine mapping_handler_init_output_fields(this, design_out, &
+       sensitivity_out, verbose_design, verbose_sensitivity, output_precision)
+    class(mapping_handler_t), intent(inout) :: this
+    type(field_t), target, intent(inout) :: design_out
+    type(field_t), target, intent(inout) :: sensitivity_out
+    logical, intent(in), optional :: verbose_design
+    logical, intent(in), optional :: verbose_sensitivity
+    integer, intent(in), optional :: output_precision
```
timfelle, line 509: You already set this in the type, so why do you eant jt as an input here?

gloopydoop: Previously the `design_brinkman_t` was responsible for the output (that's the type that reads from the JSON as well)
Now the ´mapping_handler_t´ is responsible for the output, but it's still initialized by the `design_brinkman_t` with those flags, and the `design_brinkman_t` still reads from the JSON.

The logic is that in the future we may want to map to various different coefficients, so we would have multiple mapping handlers and this give us the flexibility to initialize them in different ways.

The ones set in the type are the defaults.

sources/mapping_functions/mapping_handler.f90
```diff
@@ -122,6 +236,22 @@ subroutine mapping_handler_free(this)
...
+    if (this%outputs_initialized) then
+       call this%design_output%free()
+       call this%sensitivity_output%free()
+    end if
+    this%outputs_initialized = .false.
+    this%output_fields_set = .false.
+    this%output_precision = sp
+    nullify(this%design_out)
+    nullify(this%sensitivity_out)
+    nullify(this%coef)
```
timfelle: Missing guards for non initiated pointers
```suggestion
    if (associated(this%design_out)) nullify(this%design_out)
    if (associated(this%sensitivity_out)) nullify(this%sensitivity_out)
    if (associated(this%coef)) nullify(this%coef)
```

timfelle (approved):

PR #378, opened 2026-03-11
Mapping/heaviside continuation

AbbasMoosaV:
This PR adds a continuation_scheduler_t. Similar to neko_scratch_registry, we will have a nekotop_continuation as a global object that is initialized in the driver and it would be possible for all modules to register different parameters in there to have a continuation approach for them during the optimization loop.


sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -87,13 +89,40 @@ subroutine heaviside_mapping_init_from_json(this, json, coef)
...
+
+    ! Register continuation if multiple values
+    if (size(beta_values) > 1) then
+       ! Read optional iterations per beta step
+       call json_get_or_default(json, 'beta_iterations', beta_iter, &
+            nekotop_continuation%default_iterations)
+       call nekotop_continuation%register_parameter("beta", this%beta, &
+            beta_values, beta_iter)
+    end if
+
```
gloopydoop: I think I'll leave this one to you and Tim. But I think from a broad view, it looks pretty specific to `heaviside_mapping.f90` and I worry that for this functionality to be available for all parameters it will inflate everyone `init` significantly. Take a look in `neko_ext/json_utils_ext.f90`, surely something can be added there, some kind of `json_get_or_schedule` :sweat_smile: or something to that effect so that the changes to `heaviside_mapping.f90` essentially become a one liner. Then we can do the same for other parameters we may want to include

sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -87,13 +90,25 @@ subroutine heaviside_mapping_init_from_json(this, json, coef)
...
-    real(kind=rp) :: beta, eta
+    real(kind=rp), allocatable :: beta_values(:)
+    real(kind=rp) :: eta
+    integer :: beta_iter
 
-    call json_get(json, 'beta', beta)
     call json_get_or_default(json, 'eta', eta, 0.5_rp)
 
+    ! Read beta values from json and if that is an array also read beta_iter
+    call json_get_with_continuation(json, 'beta', beta_values, 1.0_rp,&
```
gloopydoop:
```suggestion
    call json_get_with_continuation(json, 'beta', beta_values, 1.0_rp, &
```

sources/mapping_functions/heaviside_mapping.f90
```diff
@@ -87,13 +90,25 @@ subroutine heaviside_mapping_init_from_json(this, json, coef)
...
+         beta_iter)
+
+    ! Initialize base
     call this%init_base(json, coef)
-    call this%init_from_attributes(coef, beta, eta)
+    call this%init_from_attributes(coef, beta_values(1), eta)
+
+    ! Register beta for continuation if we have an array of values for beta
```
gloopydoop: could we not pass `nekotop_continuation` into `json_get_with_continuation` and have this registering done internally in  `json_get_with_continuation`?

Then we will really have a one liner change for other parts of the code:
`call json_get(json, 'beta', beta)` -> `call json_get_with_continuation(json, 'beta', beta_values, 1.0_rp, nekotop_continuation)`

examples/rugby_ball/rugby_ball.case
```diff
@@ -87,11 +87,19 @@
...
                 {
                     "type": "PDE_filter",
                     "r": 0.01
                 },
+                {
+                    "type": "heaviside_mapping",
+                    "beta": [1.0, 10.0, 100.0],
+                    "beta_iterations": 5,
+                    "eta": 0.5,
+                },
```
gloopydoop: hmmm, maybe we should stop and think before changing the rugby ball example. The filter is there to prevent Gibbs oscillations, we **don't** want something fully black and white (even though it's stable, I would prefer if our examples emphasized the filtering)

examples/rugby_ball/rugby_ball.case
```diff
@@ -112,8 +120,9 @@
         },
         "solver": {
             "type": "mma",
-            "max_iterations": 5,
-            "tolerance": 1.0e-3,
+            "max_iterations": 15,
```
gloopydoop: these changes are good though, I think we should keep these ones 😅

gloopydoop (dismissed):
I think it looks cool, but I still think we can tighten it into a 1 line change throughout the code for each parameters

sources/mapping_functions/heaviside_mapping.f90
gloopydoop: on a second read I have one more nitpicky suggestion 😅 
In neko we have a clear distinction between `json_get` and `json_get_or_default`, which is pretty useful in forcing the user to prescribe some, but not all, parameters. This is now lost and the naming convention is strange in this function.  Intuitively the function you've written is `json_get_or_default_with_continuation` since the default is prescribed. Could we have both maybe?

timfelle (changes requested):
Running the example. I cannot see any changes to beta in the optimization_data.csv. It remain constant.

sources/mapping_functions/heaviside_mapping.f90
AbbasMoosaV: The thing is that to register a parameter in nekotop_continuation, you should send in a pointer and for example in this case it is going to be heaviside_mapping_t_object%beta.
I thought about doing that, but I guess sending a parameter to another module to be linked with another object, makes the code a bit messy and hurts the encapsulation.
I think from a software development point of view, it is better if we only use the functions in `json_utils_ext` as just helper functions to read the json file and then let the objects initializer to register their parameters in nekotop_continuation.
What do you think @timfelle?


Regarding the naming, I feel like `json_get_or_default_with_continuation` is too long for the name but if you think that would help to make things more clear, we can do that.

AbbasMoosaV:
> Running the example. I cannot see any changes to beta in the optimization_data.csv. It remain constant.

There was a little bug in `json_utils_ext`, but it is resolved now.

sources/drivers/topopt.f90
```diff
@@ -81,6 +81,9 @@ program topopt
   ! -------------------------------------------------------------------------- !
   ! Initialization of the components
 
+  ! initialize the global continuation_scheduler object (nekotop_continuation)
+  call nekotop_continuation%init(parameters)
```
timfelle, line 86: The continuation scheduler is never freed again. And you should enable this for
the topopt-user driver as well.

sources/mapping_functions/heaviside_mapping.f90
timfelle: There is a significantly better way of doing this which does not involve a new
json type object. We can make use of the registry types that Timofey implemented
on the neko side.

I can run through it but it will alter a lot of what is done here. But
essentially should allow us to trivially add any parameter to the continuation
scheme.

sources/optimizer/mma_optimizer.f90
```diff
@@ -413,6 +426,11 @@ subroutine mma_optimizer_write(this, iter, problem)
     end if
     extras(3) = this%scaling_factor
 
+    ! Continuation parameter values
+    do i = 1, n_cont
+       extras(3 + i) = nekotop_continuation%params(i)%target
+    end do
+
```
timfelle: Add this to the write_log in optimizer. We would also want this for any non-mma
optimizers.

sources/drivers/topopt.f90
timfelle: Probably better to add in the optimizer as that is the one who does the
continuation stuff.

sources/optimizer/mma_optimizer.f90
```diff
@@ -249,13 +250,7 @@ subroutine mma_optimizer_initialize(this, problem, design, simulation)
 
     ! Evaluate the problem based on the updated design
     call problem%compute(design, simulation)
-    if (present(simulation) .and. this%enable_output) then
-       call simulation%write_forward(0)
-    end if
     call problem%compute_sensitivity(design, simulation)
-    if (present(simulation) .and. this%enable_output) then
-       call simulation%write_adjoint(0)
-    end if
```
gloopydoop, line 258: why is this removed?

timfelle (approved):

PR #384, opened 2026-03-17
Brinkman dissipation naming convention

gloopydoop:
I think it's finally time to get rid of the "lube_term" :sweat_smile: 

This PR does two things:

1) Switches the naming convention from `minimum_dissipation` -> `viscous_dissipation` and `lube_term` -> `brinkman_dissipation`
2) If we want to think about it this way, there is an important prefactor of 1/Re attached to the viscous dissipation. This shouldn't change the rugby ball case so much since it was Re approx 1, but the higher Reynolds number cases will notice a difference.

If we want to treat the brinkman_dissipation as numerical parameter not a physical one, we still have the weights up our sleeve :+1:


AbbasMoosaV (commented):
This looks good.
I think the updated naming makes much more sense.
To me, adding viscosity to the Brinkman dissipation is like a weight scaling to make a physical sense out of it.
I think we can merge this in, and then I will resolve the conflicts for the naming with my  PR on continuation for objective weights.
I leave this to @timfelle if he has any comments on this.

sources/source_terms/neko_top_source_terms.f90
```diff
@@ -51,23 +52,23 @@
...
+    procedure(source_term_allocate), pointer :: adjoint_viscous_dissipation
     procedure(source_term_allocate), pointer :: adjoint_mixing_scalar
     procedure(source_term_allocate), pointer :: adjoint_scalar_convection
     procedure(source_term_allocate), pointer :: simple_brinkman
 
     ! Assign the pointers
-    adjoint_lube => adjoint_lube_source_term_allocate
-    adjoint_minimum_dissipation => &
-         adjoint_minimum_dissipation_source_term_allocate
+    adjoint_lube => adjoint_brinkman_dissipation_source_term_allocate
```
timfelle: Missing rename, still called adjoint_lube

sources/source_terms/neko_top_source_terms.f90
```diff
@@ -51,23 +52,23 @@
...
+    adjoint_viscous_dissipation => &
+         adjoint_viscous_dissipation_source_term_allocate
     adjoint_mixing_scalar => adjoint_mixing_scalar_source_term_allocate
     adjoint_scalar_convection => adjoint_scalar_convection_source_term_allocate
     simple_brinkman => simple_brinkman_source_term_allocate
 
     ! Register the source terms
     call register_source_term('adjoint_lube', adjoint_lube)
```
timfelle: Still called lube

timfelle (changes requested):

sources/source_terms/neko_top_source_terms.f90
gloopydoop: good catch :+1:

timfelle (approved):

PR #386, opened 2026-03-18
objective function over specified time windows

gloopydoop:
This PR allows for objectives to be integrated over a specified time. The idea is:
$F(x) = \int_{t_1}^{t_2} f(x) dt$
with $0 < t_1 < t_2 < T$

We can consider some function $c(t)$ which is 1 for $t_1 < t < t_2$ and zero for all other times.

So the adjoint derivation can proceed as usual with the full time integral $F(x)  \int_{0}^{T} c(t) f(x) dt$

When this is perturbed w.r.t $x$ it becomes the forcing term in the adjoint equations, so adjoint forcing terms are only active in $t_1 < t < t_2$.


sources/problem/objectives/viscous_dissipation_objective.f90
```diff
@@ -154,6 +154,7 @@ subroutine viscous_dissipation_init_json_sim(this, json, design, simulation)
     call json_get_or_default(json, "weight", weight, 1.0_rp)
     call json_get_or_default(json, "mask_name", mask_name, "")
     call json_get_or_default(json, "name", name, "Dissipation")
+    call this%init_time_window_json(json)
```
timfelle: Why did you decide to not allow a direct initialization with the time thing? Why
should it only be possible to enable through the json system?

sources/source_terms/adjoint_viscous_dissipation_source_term.f90
```diff
@@ -142,27 +142,22 @@ end subroutine adjoint_viscous_dissipation_source_term_init_from_json
...
+    real(kind=rp) :: obj_scale
     class(point_zone_t), intent(in), target :: mask
-    real(kind=rp), intent(in) :: volume
     logical :: if_mask
-
-    ! I wish you didn't need a start time and end time...
-    ! but I'm just going to set a super big number...
-    start_time = 0.0_rp
-    end_time = 100000000.0_rp
+    type(coef_t) :: coef
```
timfelle: Missing intent, please go through all the subroutines you modified and make sure
intent is properly indicated

timfelle (changes requested):

sources/source_terms/adjoint_viscous_dissipation_source_term.f90
gloopydoop: c19e041fb5d67beeac412d46c3d9a8bd4e2516c3

sources/problem/objectives/viscous_dissipation_objective.f90
gloopydoop: I've put a little helper function in case you want to set it externally instead of the json route 89b6a07f151623eb320deb925b38250b8c5999d1

sources/problem/base_functional.f90
```diff
@@ -171,6 +182,24 @@ subroutine functional_init_json_sim(this, json, design, simulation)
...
+
+  !> Set the active time window for this functional externally.
+  subroutine functional_set_time_window(this, start_time, end_time)
+    class(base_functional_t), intent(inout) :: this
+    real(kind=rp), intent(in) :: start_time
+    real(kind=rp), intent(in) :: end_time
+
+    this%start_time = start_time
+    this%end_time = end_time
+  end subroutine functional_set_time_window
```
timfelle: This will not alter the source terms and other components, which have already
been initialized inside the objects.

sources/problem/objectives/viscous_dissipation_objective.f90
timfelle: Please read start and end time, and pass them to the `init_from_attributes`.
Your helper does not alter the `adjoint_forcing` once the objective is
initialized.
That approach is how every single other component functions in Neko and our code.

timfelle (changes requested):

timfelle (changes requested):
Please follow the pattern used everywhere else.

Add parameters to init from attributes, as was already requested in a previous comment

sources/problem/objectives/viscous_dissipation_objective.f90
```diff
@@ -165,17 +169,24 @@ end subroutine viscous_dissipation_init_json_sim
...
     character(len=*), intent(in) :: name
     character(len=*), intent(in) :: mask_name
+    real(kind=rp), intent(in) :: start_time
+    real(kind=rp), intent(in) :: end_time
+
     type(adjoint_viscous_dissipation_source_term_t) :: adjoint_forcing
 
     call this%init_base(name, design%size(), weight, mask_name)
+    this%start_time = start_time
+    this%end_time = end_time
```
timfelle: They live in the base class. Why are the base class never initialized with them?

sources/source_terms/adjoint_viscous_dissipation_source_term.f90
```diff
@@ -142,27 +142,23 @@ end subroutine adjoint_viscous_dissipation_source_term_init_from_json
...
     real(kind=rp), intent(in) :: volume
-    logical :: if_mask
+    real(kind=rp), intent(in) :: start_time
+    real(kind=rp), intent(in) :: end_time
 
-    ! I wish you didn't need a start time and end time...
-    ! but I'm just going to set a super big number...
-    start_time = 0.0_rp
-    end_time = 100000000.0_rp
+    type(field_list_t) :: fields
```
timfelle, line 161: The fields are never freed

sources/problem/base_functional.f90
```diff
@@ -72,6 +73,10 @@ module base_functional
      logical :: has_mask = .false.
      !> A mask for where the objective function is evaluated
      class(point_zone_t), pointer :: mask => null()
+     !> Time window start for accumulation
+     real(kind=rp) :: start_time = 0.0_rp
+     !> Time window end for accumulation
+     real(kind=rp) :: end_time = huge(0.0_rp)
```
timfelle, line 79: These are never initialized. Please make sure a user can utilize these.

timfelle (changes requested):

timfelle (approved):

PR #391, opened 2026-03-25
Field names in mapping handler

gloopydoop:
This PR has a few quality of life features in the mapping handler
- `fld_file_output_t` -> `field_output_t` allowing different output formats
- names for various stages in the mapping handler

adding this to the rugby ball example
```
    "optimization": {
        "design": {
            "type": "brinkman",
            "dealias": false,
            "output_format": "vtkhdf",
            "verbose_design": true,
            "verbose_sensitivity": true,
```

should yield

image

### DIFF
--- a/documentation/pages/user_guide/design_types.md
+++ b/documentation/pages/user_guide/design_types.md
@@ -37,9 +37,13 @@ Brinkman source-term coupling.
 | `dealias` | bool | No | `true` | Controls dealiasing for Brinkman source terms |
 | `verbose_design` | bool | No | `false` | Write all forward mapping stages |
 | `verbose_sensitivity` | bool | No | `false` | Write all backward sensitivity stages |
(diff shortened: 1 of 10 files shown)


timfelle (changes requested):
What happens to the name when you have a mapping object for the design and a mapping object for a constraint or an objective? How does one distinguish between these? From what I can see, no matter what you attach a mapping to it is treated as a "design" and the same file will be overwritten, with no way of knowing which mapping actually did the overwritting last.

gloopydoop:
> What happens to the name when you have a mapping object for the design and a mapping object for a constraint or an objective? How does one distinguish between these? From what I can see, no matter what you attach a mapping to it is treated as a "design" and the same file will be overwritten, with no way of knowing which mapping actually did the overwritting last.

good point, ffa5ade70d1e2aec3c21b74e0955088ab5b91d81

timfelle (approved):

PR #398, opened 2026-03-31
Add PDE_filter radius as a continuation parameter.

AbbasMoosaV:
This PR extends the continuation framework to support the filter radius (r) in the PDE filter.
Note that we store the physical filter radius `this%r` as defined in the case file; internally, the module uses r / ( 2 ∗ s q r t ( 3 ) ) in all the calculations to ensure consistency with the equivalent PDE/Helmholtz filter length scale.


gloopydoop (approved):
Looks good with the addition of #401

PR #405, opened 2026-04-07
Only one rank cleaning checkpoints

gloopydoop:
This PR ensure just a single rank does the checkpoint cleaning


gloopydoop:
allready fixed somewhere else

PR #478, opened 2026-08-17
Fix premature removal of files during rsync failure

timfelle:
Implement error handling to prevent file deletion when rsync encounters an error, ensuring that important logs are retained for troubleshooting.


gloopydoop (approved):
👍  nice

